PR #22, opened 2019-08-14
Change error parameter name to reason

### LINKED ISSUE #21: Possible typo: reason instead of error
The code in this page:

https://metamask.github.io/metamask-docs/API_Reference/Ethereum_Provider#ethereum.enable()

```
Example usage (ES6), assuming async function:

try {
  const accounts = await ethereum.enable()
  // You now have an array of accounts!
  // Currently only ever one:
  // ['0xFDEa65C8e26263F6d9A1B5de9555D2931A33b825']

} catch (error) {
  // Handle error. Likely the user rejected the login:
  console.log(reason === "User rejected provider access")
}
```

logs an error: `reason is not defined`

`reason` might have been mistaken which should've been `error`

sebastiantf:
Fixes #21

### DIFF
--- a/03_API_Reference/01_Ethereum_Provider.md
+++ b/03_API_Reference/01_Ethereum_Provider.md
@@ -54,7 +54,7 @@ try {
   // Currently only ever one:
   // ['0xFDEa65C8e26263F6d9A1B5de9555D2931A33b825']
 
-} catch (error) {
+} catch (reason) {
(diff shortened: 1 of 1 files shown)


Gudahtt:
Thanks! I think we'd rather go the other way actually - use `error` in both places instead of `reason`. I've also removed the strange string comparison, and made this same change in a few other places in thid document.

PR #36, opened 2019-11-07
Update documentation for December 16 changes

rekmarks:
In preparation for the December 16 provider updates, I reviewed our existing Provider API documentation and added the new API to the end of the page.

In the process of doing this, I noticed that our existing documentation of `ethereum.send` was out of sync with our implementation. Our [current implementation of this method](https://github.com/MetaMask/metamask-inpage-provider/blob/master/index.js#L98-L107) is either (1) an alias of `sendAsync(requestObject, callback)`, or (2) a route to four synchronous methods using `send(requestObject)`. (2) is implemented [here](https://github.com/MetaMask/metamask-inpage-provider/blob/master/index.js#L121-L156). At the moment, `send` is [documented as Promise-returning](https://metamask.github.io/metamask-docs/API_Reference/Ethereum_Provider#ethereum.send(options)). Attempting to use it in the manner described by the documentation throws an error.

This PR changes the documentation of the current state of `send` as an alias of `sendAsync`. It does not add documentation for (2), as it will soon be removed. It also adds a notice about the breaking changes on December 16, and adds documentation for the new API.

This PR will be followed up with other PRs as we approach the launch of the new API during the week of November 25, and then the removal of the current API on December 16.

### DIFF
--- a/03_API_Reference/01_Ethereum_Provider.md
+++ b/03_API_Reference/01_Ethereum_Provider.md
@@ -1,35 +1,64 @@
-# API Reference
+# Ethereum Provider API
 
-MetaMask injects a global API into websites visited by its users at `window.ethereum` (Also available at `window.web3.currentProvider` for legacy reasons). This API allows websites to request user login, load data from blockchains the user has a connection to, and suggest the user sign messages and transactions. You can use this API to detect the user of a web3 browser.
+MetaMask injects a global API into websites visited by its users at `window.ethereum`.
+This API allows websites to request user login, load data from blockchains the user has a connection to, and suggest that the user sign messages and transactions.
+You can use this API to detect the user of an Ethereum browser.
 
 ```javascript
-if (typeof window.ethereum !== 'undefined'
-|| (typeof window.web3 !== 'undefined')) {
+if (typeof window.ethereum !== 'undefined') {
 
-  // Web3 browser user detected. You can now use the provider.
-  const provider = window['ethereum'] || window.web3.currentProvider
+  // Ethereum user detected. You can now use the provider.
+  const provider = window['ethereum']
 }
 ```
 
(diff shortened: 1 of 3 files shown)


danfinlay:
Approved, but also keep in mind we have that new docs site ready to transition to our org soon. We will need to migrate these changes to it.

03_API_Reference/01_Ethereum_Provider.md
```diff
@@ -92,11 +131,126 @@ params: [{
...
+This API will become available during the week of **November 25, 2019**.
+It will be the only API supported starting **December 16, 2019.**
+If you want examples of how to setup your dapp using the new API, you can check out
+[this gist](https://gist.github.com/rekmarks/d318677c8fc89e5f7a2f526e00a0768a).
+
+## Methods
+
+### ethereum.send('eth_requestAccounts')
+
+#### ethereum.enable() (Deprecated)
```
Gudahtt: It wasn't immediately clear reading this that `ethereum.enable` was shown here because it's equivalent to `ethereum.send('eth_requestAccounts')`. Maybe referring to it as an alias/alternative/equivalent/synonym, or having a sentence to contextualize this would be helpful.

rekmarks:
Whoever sees this, please do not delete the `new-provider` branch.

03_API_Reference/01_Ethereum_Provider.md
```diff
@@ -105,13 +259,45 @@ ethereum.send({
...
+'2': Morden Test network
+'3': Ropsten Test Network
+'4': Rinkeby Test Network
+'5': Goerli Test Network
+'42': Kovan Test Network
+```
+
+### ethereum.sendAsync(options, callback) (Deprecated)
+
+_We strongly discourage the use of this method, which may be removed in the future._
```
Gudahtt: Ideally we would link to the recommended alternative for each deprecated item. Maybe a line following each of these "We strongly discourage" sentences would be the right place for that?

03_API_Reference/01_Ethereum_Provider.md
```diff
@@ -133,31 +319,31 @@ ethereum.sendAsync({
 })
 ```
 
-### ethereum.autoRefreshOnNetworkChange
+## Properties
 
-When the network is changed, MetaMask will reload any pages that have made requests to the provider. This automatic reload behavior will be removed in a future release of MetaMask, but in the meantime it can be disabled with this flag.
+Useful for knowing whether `window.ethereum` is MetaMask, but not much else.
```
Gudahtt: Was this left here by mistake? It sounds like it belongs under the `isMetaMask` header

PR #116, opened 2020-06-18
Miscellanous Introduction improvements

Gudahtt:
The wording has been improved in a few places in the introduction. The last section has been removed as well ("Ethereum Applications"), because I found it to be more confusing than helpful, but I didn't know how to improve it.

### DIFF
--- a/docs/guide/README.md
+++ b/docs/guide/README.md
@@ -9,7 +9,7 @@ Welcome to MetaMask’s Developer Documentation. This documentation is for learn
 
 ## Why Metamask
 
-MetaMask was created out of the needs of creating more secure and usable Ethereum-based web sites. In particular, it handles account management and connecting the user to the blockchain.
+MetaMask was created to meet the needs of secure and usable Ethereum-based web sites. In particular, it handles account management and connecting the user to the blockchain.
 
 - [Get started here](/guide/getting-started.html)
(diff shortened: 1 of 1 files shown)


docs/guide/README.md
```diff
@@ -9,7 +9,7 @@ Welcome to MetaMask’s Developer Documentation. This documentation is for learn
 
 ## Why Metamask
 
-MetaMask was created out of the needs of creating more secure and usable Ethereum-based web sites. In particular, it handles account management and connecting the user to the blockchain.
+MetaMask was created to meet the needs of secure and usable Ethereum-based web sites. In particular, it handles account management and connecting the user to the blockchain.
```
brad-decker, line 12: The change to this first line seems to change the point from 'MetaMask creates a more secure connection to sites' to 'MetaMask provides authors a secure way to interact with users'. Is that the intent here?

Gudahtt: I had difficulty interpreting the original sentence, to be honest. I remain unsure what "creates a more secure connection to sites" would mean.

brad-decker: I think I extrapolated a little too hard there... The original doesn't mention connections at all. It implies MetaMask creates more secure _and usable_ *web sites*. In that light, your rewording aligns much more closely with what MetaMask actually does.

PR #104, opened 2020-05-08
WIP: Document new Provider API

rekmarks:
WIP

### New/updated files

- https://github.com/MetaMask/metamask-docs/blob/1193-new/docs/guide/ethereum-provider.md

PR #123, opened 2020-06-26
v8 Provider Documentation

rekmarks:
First of a number of v8 PRs using the `v8` branch as a base. Compare #104 for the finished product. We're splitting it up in order to make review more manageable.

### Assumptions
- This assumes that the "Experimental API" page will be removed, and its content moved into the Provider and RPC API pages.

### DIFF
--- a/docs/guide/ethereum-provider.md
+++ b/docs/guide/ethereum-provider.md
@@ -131,6 +131,29 @@ The value of this property can change at any time, and should not be exclusively
 
 A hexadecimal string representing the current chain ID.
 
+### ethereum.autoRefreshOnNetworkChange
+
+::: warning
+The value of this property will only affect MetaMask's behavior if `window.web3` is accessed during the page lifecycle.
+When `window.web3` [is removed](#window-web3-removal), either the effects of this property will change, or it will be removed.
+
+As the consumer of this API, it is your responsbility to handle chain changes using the [`chainChanged` event](#chainChanged).
+We recommend reloading the page on `chainChange` unless you have good reason not to.
+:::
+
+By default, this property is `true`.
+
+If this property is truthy, MetaMask will reload the page in the following cases:
+
+- When the connected chain (network) changes, if `window.web3` has been accessed during the page lifecycle
+- When `window.web3` is accessed, if the connected chain (network) has changed during the page lifecycle
+
+To disable this behavior, set this property to `false` immediately after detecting the provider:
+
+```javascript
+ethereum.autoRefreshOnNetworkChange = false;
+```
+
 ## Methods
 
 ### ethereum.isConnected()
@@ -395,29 +418,6 @@ MetaMask only supported this API before the provider API was standardized via [E
 Because of this, you may find web3 sites that use this API, or other providers that implement it.
 
 ## Legacy Properties
-
-### ethereum.autoRefreshOnNetworkChange (DEPRECATED)
-
-::: warning
-The value of this property will only affect MetaMask's behavior if `window.web3` is accessed during the page lifecycle.
-When `window.web3` [is removed](#window-web3-removal), either the effects of this property will change, or it will be removed.
-
-As the consumer of this API, it is your responsbility to handle chain changes using the [`chainChanged` event](#chainChanged).
-We recommend reloading the page on `chainChange` unless you have good reason not to.
-:::
-
-By default, this property is `true`.
-
-If this property is truthy, MetaMask will reload the page in the following cases:
-
-- When the connected chain (network) changes, if `window.web3` has been accessed during the page lifecycle
-- When `window.web3` is accessed, if the connected chain (network) has changed during the page lifecycle
-
-To disable this behavior, set this property to `false` immediately after detecting the provider:
-
-```javascript
-ethereum.autoRefreshOnNetworkChange = false;
-```
 
 ### ethereum.networkVersion (DEPRECATED)
 
--- a/docs/guide/ethereum-provider.md
+++ b/docs/guide/ethereum-provider.md
@@ -35,9 +35,11 @@ We recommend that all web3 site developers read the [Basic Usage section](#basic
 ## Upcoming Breaking Changes
 
 ::: tip
-These changes are _upcoming._ Follow [this GitHub issue](https://github.com/MetaMask/metamask-extension/issues/8077) for details.
+These changes are _upcoming._ Follow [this GitHub issue](https://github.com/MetaMask/metamask-extension/issues/8077) for updates.
 
-If you are new to using the provider, or use [ethers](https://www.npmjs.com/package/ethers), you do not have to worry about these changes, and can skip ahead [to the next section](#api).
+All consumers of this API will be affected by the `eth_chainId` bug fix (see [next subsection](#window-ethereum-api-changes)).
+
+Otherwise, if you are new to using the provider or use [ethers](https://www.npmjs.com/package/ethers), you do not have to worry about these changes, and can skip ahead [to the next section](#api).
 :::
 
 ### `window.ethereum` API Changes
@@ -47,9 +49,9 @@ In **Q3 2020** (date TBD), we are introducing some breaking changes to this API,
 
 At that time, we will:
 
-- Stop emitting `chainIdChanged`, and instead emit `chainChanged`
 - Ensure that chain IDs returned by `eth_chainId` are **not** 0-prefixed
   - For example, instead of `0x01`, we will always return `0x1`, wherever the chain ID is returned or accessible.
+- Stop emitting `chainIdChanged`, and instead emit `chainChanged`
 - Remove the following experimental methods:
   - `ethereum._metamask.isEnabled`
   - `ethereum._metamask.isApproved`
@@ -371,8 +373,8 @@ ethereum._metamask.isUnlocked(): Promise<boolean>;
 ```
 
 This method returns a `Promise` that resolves to a `boolean` indicating if MetaMask is unlocked by the user.
-This is useful for knowing if MetaMask is unlocked in order to provide meaningful instructions to the user during onboarding.
-Note that this does not indicate if the user has exposed any accounts.
+MetaMask must be unlocked in order to perform any operation involving user accounts.
+Note that this method does not indicate if the user has exposed any accounts to the caller.
 
 ## Legacy API
 
@@ -380,8 +382,9 @@ Note that this does not indicate if the user has exposed any accounts.
 You should **never** rely on any of these methods, properties, or events in practice.
 :::
 
-Historically, the Ethereum provider API has been a mess.
-Here, we document the remains of this mess in case you see it used in the wild and wonder what's going on.
+This section documents our legacy provider API.
+MetaMask only supported this API before the provider API was standardized via [EIP-1193](https://eips.ethereum.org/EIPS/eip-1193) in 2020.
+Because of this, you may find web3 sites that use this API, or other providers that implement it.
 
 ## Legacy Properties
 
--- a/docs/guide/ethereum-provider.md
+++ b/docs/guide/ethereum-provider.md
@@ -55,7 +55,7 @@ At that time, we will:
   - `ethereum._metamask.isApproved`
 
 These changes _may_ break your website.
-Please read our [migration guide](about:blank) for more details.
+Please read our [migration guide](./provider-migration.html) for more details.
 
 ### `window.web3` Removal
 
@@ -69,7 +69,7 @@ In **Q4 2020** (date TBD), we will:
 - Remove the `ethereum.autoRefreshOnNetworkChange` property
 
 If you rely on the `window.web3` object currently injected by MetaMask, these changes _will_ break your website.
-Please read our [migration guide](about:blank) for more details.
+Please read our [migration guide](./provider-migration.html) for more details.
 
 ## Basic Usage
 
--- a/docs/guide/ethereum-provider.md
+++ b/docs/guide/ethereum-provider.md
@@ -283,7 +283,8 @@ ethereum.on('message', handler: (message: ProviderMessage) => void);
 The MetaMask provider emits this event when it receives some message that the consumer should be notified of.
 The kind of message is identified by the `type` string.
 
-One prominent example of such messages include updates from RPC subscriptions, such as `eth_subscribe`. For such messages, the message `type` will be `eth_subscription`.
+RPC subscription updates are a common use case for the `message` event.
+For example, if you create a subscription using `eth_subscribe`, each subscription update will be emitted as a `message` event with a `type` of `eth_subscription`.
 
 ## Errors
 
--- a/docs/guide/ethereum-provider.md
+++ b/docs/guide/ethereum-provider.md
@@ -48,7 +48,7 @@ In **Q3 2020** (date TBD), we are introducing some breaking changes to this API,
 At that time, we will:
 
 - Stop emitting `chainIdChanged`, and instead emit `chainChanged`
-- Ensure that all `chainId` values are **not** 0-prefixed
+- Ensure that chain IDs returned by `eth_chainId` are **not** 0-prefixed
   - For example, instead of `0x01`, we will always return `0x1`, wherever the chain ID is returned or accessible.
 - Remove the following experimental methods:
   - `ethereum._metamask.isEnabled`
--- a/docs/guide/ethereum-provider.md
+++ b/docs/guide/ethereum-provider.md
@@ -238,7 +238,9 @@ ethereum.on('accountsChanged', handler: (accounts: Array<string>) => void);
 ```
 
 The MetaMask provider emits this event whenever the return value of the `eth_accounts` RPC method changes.
-`eth_accounts` returns an array that is either empty or contains a single account address that the user decided to expose to the requesting domain (in the case of a website, identified by the _hostname_ of its URL).
+`eth_accounts` returns an array that is either empty or contains a single account address.
+The returned address, if any, is the address of the most recently used account that the caller is permitted to access.
+Callers are identified by their URL _origin_, which means that all sites with the same origin share the same permissions.
 
 This means that `accountsChanged` will be emitted whenever the user's exposed account address changes.
 
--- a/docs/guide/ethereum-provider.md
+++ b/docs/guide/ethereum-provider.md
@@ -336,6 +336,8 @@ We expose some experimental, MetaMask-specific methods under the `ethereum._meta
 
 ::: danger DANGER
 This will be removed in **Q3 2020**.
+
+Please see the [Using the Provider section](#using-the-provider) for the recommended way of keeping track of user accounts.
 :::
 
 ```typescript
@@ -348,6 +350,8 @@ This method returns a `Promise` that resolves to a `boolean` indicating if the c
 
 ::: danger DANGER
 This will be removed in **Q3 2020**.
+
+Please see the [Using the Provider section](#using-the-provider) for the recommended way of keeping track of user accounts.
 :::
 
 ```typescript
--- a/docs/guide/ethereum-provider.md
+++ b/docs/guide/ethereum-provider.md
@@ -121,7 +121,7 @@ Consult [chainid.network](https://chainid.network) for more.
 This property is not guaranteed to be correct for all providers. Non-MetaMask providers may also set this property to `true`.
 :::
 
-`true` if the user has MetaMask installed, some falsy value otherwise.
+`true` if the user has MetaMask installed.
 
 ### ethereum.chainId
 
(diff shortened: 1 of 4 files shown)


docs/guide/ethereum-provider.md
```diff
@@ -1,103 +1,164 @@
...
-### ethereum.isMetaMask
+At that time, we will:
 
-Returns `true` or `false`, representing whether the user has MetaMask installed.
+- Stop emitting `chainIdChanged`, and instead emit `chainChanged`
+- Ensure that all `chainId` values are **not** 0-prefixed
+  - For example, instead of `0x01`, we will always return `0x1`, wherever the chain ID is returned or accessible.
+- Remove the following experimental methods:
+  - `ethereum._metamask.isEnabled`
+  - `ethereum._metamask.isApproved`
```
Gudahtt, line 55: .... does this exist? 😕 

I don't think we're removing this. I don't think it exists.

Gudahtt: Oh, it exists but it's not in prod. Huh. I don't think we need to leave this in for a whole deprecation period if it hasn't made it to prod yet.

docs/guide/ethereum-provider.md
```diff
@@ -1,103 +1,164 @@
...
 
-### ethereum.isMetaMask
+At that time, we will:
 
-Returns `true` or `false`, representing whether the user has MetaMask installed.
+- Stop emitting `chainIdChanged`, and instead emit `chainChanged`
+- Ensure that all `chainId` values are **not** 0-prefixed
+  - For example, instead of `0x01`, we will always return `0x1`, wherever the chain ID is returned or accessible.
+- Remove the following experimental methods:
+  - `ethereum._metamask.isEnabled`
```
Gudahtt, line 54: Why are we removing this? What would recommend as its replacement?

docs/guide/ethereum-provider.md
```diff
@@ -1,103 +1,164 @@
...
+  - `ethereum._metamask.isApproved`
 
-### ethereum.autoRefreshOnNetworkChange
+These changes _may_ break your website.
+Please read our [migration guide](about:blank) for more details.
```
Gudahtt, line 58: >`[migration guide](about:blank)`

😅 

I mean... this seems a little bit rude.

docs/guide/ethereum-provider.md
rekmarks: wat.

rekmarks: You mean `ethereum._metamask` and its properties, `isEnabled` and `isApproved`?

They do exist in prod, they're just defined in the extension.

rekmarks: I was questioning my sanity there for a second 😅 

image

docs/guide/ethereum-provider.md
```diff
@@ -1,103 +1,164 @@
...
-## Properties
+In **Q3 2020** (date TBD), we are introducing some breaking changes to this API, which we encourage you to
+[read more about here](https://medium.com/metamask/breaking-changes-to-the-metamask-inpage-provider-b4dde069dd0a).
 
-### ethereum.isMetaMask
+At that time, we will:
 
-Returns `true` or `false`, representing whether the user has MetaMask installed.
+- Stop emitting `chainIdChanged`, and instead emit `chainChanged`
+- Ensure that all `chainId` values are **not** 0-prefixed
```
Gudahtt, line 51: We should be as specific as we can here. Does this only effect `eth_chainId`? I know that `networkChanged` and `chainChanged`/`chainIdChanged` aren't effected.

I wonder if any data returned by other methods has this incorrectly-prefixed `chainId` in the metadata somewhere 🤔 e.g. `txMeta`

docs/guide/ethereum-provider.md
```diff
@@ -1,103 +1,164 @@
...
 
-## List of Chain and Network IDs
+## Upcoming Breaking Changes
 
 ::: tip
-The chain and network IDs of these chains are the same.
+These changes are _upcoming._ Follow [this GitHub issue](https://github.com/MetaMask/metamask-extension/issues/8077) for details.
 
-Network IDs are decimal strings, while chain IDs are hexadecimal strings.
+If you are new to using the provider, or use [ethers](https://www.npmjs.com/package/ethers), you do not have to worry about these changes, and can skip ahead [to the next section](#api).
```
Gudahtt, line 40: >or use ethers

The `chainId` breaking change might still applies for people using `ethers`.

docs/guide/ethereum-provider.md
```diff
@@ -1,103 +1,164 @@
...
+:::
 
-```javascript
-ethereum.autoRefreshOnNetworkChange = false;
-```
+In **Q4 2020** (date TBD), we will:
 
-This can be toggled on or off at any time.
+- Stop injecting `window.web3` into web pages
+- Remove the `ethereum.autoRefreshOnNetworkChange` property
```
Gudahtt, line 69: Are we definitely removing this? I don't think we should mention this until we're prepared to commit to removing this, and we have a plan on how to replace it (e.g. no more automatic refreshes, or a replacement property, etc.)

This leaves me confused about what I'm supposed to do about this removal. Our current error messages strongly advise that all dapps set this property after all.

docs/guide/ethereum-provider.md
Gudahtt: 🤦 I was getting this confused for the top-level methods, like `isConnected`. This makes more sense.

Gudahtt: I was referencing the older version of the inpage provider, and `createStandardProvider`. I forgot there was additional stuff added in `inpage.js` 🤦

rekmarks: https://youtu.be/NzWizk3e4MM

docs/guide/ethereum-provider.md
```diff
@@ -1,103 +1,164 @@
...
+The snippet at the top of this page is sufficient for detecting the provider.
+You can learn how to accomplish the other two by reviewing the snippet in the [Using the Provider section](#using-the-provider).
+
+Although any Ethereum operation can be performed via the provider API, most developers use a convenience library with higher-level abstractions.
+The most common such libraries are [ethers](https://www.npmjs.com/package/ethers) and [web3.js](https://www.npmjs.com/package/web3).
+
+Even if you use a convenience library, you will likely still need to detect the provider before initializing the library.
+Other than that, you can generally learn everything you need to know from the documentation of those libraries, without reading this lower-level API.
+
+However, for developers of convenience libraries, and for developers who would like to use features that are not yet supported by their favorite libraries, knowledge of the provider API is essential. Read on for more details.
```
Gudahtt, line 91: I'm not a fan of how this is framed 🤔 This seems like we're discouraging most devs from using the provider directly, and I don't think we should do that. We should reference convenience libraries, but we don't need to describe them as "high-level" and us as "low-level", and we don't need to make broad generalizations about who might want to use us directly.

Any dev who trusts the EIP-1193 standard to remain stable more than they trust `web3` or `ethers` to remain stable might want to use the provider directly. Or anyone that wants to just use fewer dependencies.

Gudahtt: I tried and failed to find a way to re-word this to avoid those problems. Maybe this whole section could be omitted from this page. We already mention convenience libraries in the 'Getting Started' guide.

docs/guide/ethereum-provider.md
```diff
@@ -112,83 +173,359 @@ params: [
...
+Once `disconnect` has been emitted, MetaMask will not accept any new requests until until `connect` is emitted, which may require reloading the page.
+
+### accountsChanged
+
+```typescript
+ethereum.on('accountsChanged', handler: (accounts: Array<string>) => void);
+```
+
+The MetaMask provider emits this event whenever the return value of the `eth_accounts` RPC method changes.
+`eth_accounts` returns an array that is either empty or contains a single account address that the user decided to expose to the requesting domain (in the case of a website, identified by the _hostname_ of its URL).
```
Gudahtt, line 240: I'm not sure this is correct exactly; we assign permissions by origin, not by hostname. We could probably say a little more about which account we're returning too (e.g. that it's the most recently used account).
Maybe something like this?
```suggestion
`eth_accounts` returns an array that is either empty or contains a single account address. The account address returned (if any) is the most recently used account that the requesting site has permission to access. Permissions are granted to _origins_, so all sites on the same origin share the same permissions.
```

docs/guide/ethereum-provider.md
```diff
@@ -112,83 +173,359 @@ params: [
...
+**NOTE:** See the [Chain IDs section](#chain-ids) for important information about the MetaMask provider's chain IDs.
+:::
+
+```typescript
+ethereum.on('chainChanged', handler: (chainId: string) => void);
+```
+
+The MetaMask provider emits this event when the currently connected chain changes.
+
+All RPC requests are be submitted to the currently connected chain.
```
Gudahtt, line 260:
```suggestion
All RPC requests are submitted to the currently connected chain.
```

docs/guide/ethereum-provider.md
```diff
@@ -112,83 +173,359 @@ params: [
...
+  data: unknown;
+}
+
+ethereum.on('message', handler: (message: ProviderMessage) => void);
+```
+
+The MetaMask provider emits this event when it receives some message that the consumer should be notified of.
+The kind of message is identified by the `type` string.
+
+One prominent example of such messages include updates from RPC subscriptions, such as `eth_subscribe`. For such messages, the message `type` will be `eth_subscription`.
```
Gudahtt, line 283: The grammar for this seemed a bit off? It seemed like it was suggesting that `eth_subscription` was the `type` of _any_ "RPC subscriptions" rather than `eth_subscribe` specifically, which is obviously what was meant.
```suggestion
One prominent example is `eth_subscribe` updates. Each update for an `eth_subscribe` RPC subscription will be a message with a `type` of `eth_subscription`.
```

docs/guide/ethereum-provider.md
```diff
@@ -112,83 +173,359 @@ params: [
...
+This method returns a `boolean` indicating if the caller has access to user accounts.
+
+### ethereum.\_metamask.isUnlocked()
+
+```typescript
+ethereum._metamask.isUnlocked(): Promise<boolean>;
+```
+
+This method returns a `Promise` that resolves to a `boolean` indicating if MetaMask is unlocked by the user.
+This is useful for knowing if MetaMask is unlocked in order to provide meaningful instructions to the user during onboarding.
```
Gudahtt, line 366: During onboarding specifically? This seems useful generally all of the time, not just during onboarding. The dapp can't do much if the extension is locked.

docs/guide/ethereum-provider.md
```diff
@@ -112,83 +173,359 @@ params: [
...
+## Legacy API
 
 ::: warning
-Use `ethereum.sendAsync()` instead.
+You should **never** rely on any of these methods, properties, or events in practice.
 :::
 
-See `ethereum.sendAsync`, directly below.
+Historically, the Ethereum provider API has been a mess.
+Here, we document the remains of this mess in case you see it used in the wild and wonder what's going on.
```
Gudahtt, line 376: 😅 Maybe we should try to tone down the bitterness. Our users don't need to know how difficult a journey this has been.

docs/guide/ethereum-provider.md
```diff
@@ -112,83 +173,359 @@ params: [
...
 
-### ethereum.on(eventName, callback)
+## Legacy Properties
 
-The provider supports listening for some events:
+### ethereum.autoRefreshOnNetworkChange (TO BE REMOVED)
```
Gudahtt, line 380: Same note here re: removal - it seems premature.

docs/guide/ethereum-provider.md
```diff
@@ -112,83 +173,359 @@ params: [
...
+When `window.web3` is removed in **Q4 2020**, this property will also be removed.
 
-#### Example
+If you don't access `window.web3`, the value of this property will not affect the behavior of your application or MetaMask.
+
+As the consumer of this API, it is your responsbility to handle chain changes using the [`chainChanged` event](#chainChanged).
+We recommend reloading the page on `chainChange` or using [ethers](https://www.npmjs.com/package/ethers), which will do so for you.
+:::
+
+By default, this property is `true`.
```
Gudahtt, line 391: This isn't quite true. By default this isn't set at all, and we print a console warning asking the dapp to set it.

docs/guide/ethereum-provider.md
```diff
@@ -112,83 +173,359 @@ params: [
...
+
+```typescript
+ethereum.send(
+  methodOrPayload: string | JsonRpcRequest,
+  paramsOrCallback: Array<unknown> | JsonRpcCallback,
+): Promise<JsonRpcResponse> | unknown;
+```
+
+This method is mostly a Promise-based `sendAsync` with `method` and `params` instead of a payload as arguments.
+However, it behaves in unexpected ways and should be avoided at all costs.
```
Gudahtt, line 483: We should probably be a little more specific about how it behaves, embarrassing as it may be. Some poor soul debugging a dapp might thank us for it someday.

docs/guide/ethereum-provider.md
rekmarks: lmao

Whoops, that's what I get for not labeling my TODOs. I made a placeholder for this for now and we can handle it in a separate PR.

c6a78b2

docs/guide/ethereum-provider.md
rekmarks: > Why are we removing this?

Several reasons!

- It's sync, but the state is referring to is async
- It refers to a concept called "enabled", which we've done away with
- Its purpose is to keep track of whether accounts are available, but there's `eth_accounts` and `accountsChanged` for that

> What would we recommend as its replacement?

The snippet in "Using the Provider", which is what I advise in 8b8a693

docs/guide/ethereum-provider.md
rekmarks: Good point, although I'm not sure where else it may appear. The data returned by e.g. `eth_getTransactionReceipt` and `eth_getTransactionByHash` does not contain the `chainId` (just tested).

docs/guide/ethereum-provider.md
rekmarks: > Our current error messages strongly advise that all dapps set this property after all.

For the time being, yes, but given that we're removing the object whose behavior the property controls, I don't see how we can keep the property around. I think it'll be confusing if it starts to control the behavior of `window.ethereum` once `window.web3` is removed.

What to do instead can be mentioned in the migration guide. We can have another go at discussing auto-reload today.

docs/guide/ethereum-provider.md
rekmarks: How about this?
------------------------------------
> The provider API is all you need to create a full-featured web3 application.
> 
> That said, many developers use a convenience library, such as [ethers](https://www.npmjs.com/package/ethers) and [web3.js](https://www.npmjs.com/package/web3), instead of using the provider directly.
> If you are in need of higher-level abstractions than those provided by this API, we recommend that you seek out a convenience library.

docs/guide/ethereum-provider.md
Gudahtt: It doesn't seem quite accurate to frame the automatic reload as being something the `window.web3` property _does_. I don't see this property as "controlling the behavior of `window.web3`".

I agree that it is confusing to say that we're going to remove `window.web3` without specifying how the automatic reload behavior will change. But still - better to leave the information out if we're not sure yet. Saying we'll remove the property and then deciding to keep it would be _even more_ confusing after all.

I think it is a legitimate possibility that we'll keep the property if we're keeping the automatic reload behavior itself (even if the behavior does change somewhat). We can rename it, but we'd keep the old name around for backwards-compatibility, because... why not. The only scenario where we'd want to completely remove it is if we don't do automatic page reloading anymore at all. Which... seems likely? But we should finalize that decision first in any case. Hopefully we can do that soon.

docs/guide/ethereum-provider.md
```diff
@@ -112,83 +173,359 @@ params: [
...
-#### Example
+```javascript
+ethereum.on('accountsChanged', (accounts) => {
+  // Handle the new accounts, or lack thereof.
+  // "accounts" will always be an array, but it can be empty.
+});
+
+ethereum.on('chainChanged', (chainId) => {
+  // Handle the new chain.
+  // We recommend reloading the page.
```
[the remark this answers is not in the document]
rekmarks: e93e65d

docs/guide/ethereum-provider.md
Gudahtt: That seems fine! We can always improve it later if we think of a better way to frame it.

docs/guide/ethereum-provider.md
```diff
@@ -112,83 +173,359 @@ params: [
...
+```typescript
+interface ConnectInfo {
+  chainId: string;
+}
+
+ethereum.on('connect', handler: (connectInfo: ConnectInfo) => void);
+```
+
+The MetaMask provider emits this event when it first becomes able to submit RPC requests to a chain.
+In general, you can assume that the MetaMask provider is connected and has emitted this event by the time you are able to reference it.
```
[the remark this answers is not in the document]
Gudahtt: I would suggest that for the time being, we not mention this weird behavior. We should recommend that users use `isConnected()` in conjunction with this event handler, so they can check whether the extension is _already_ connected upon page load, and can register the handler to deal with updates.

Though... why is `isConnected()` deprecated as well 🤦‍♂️. I just noticed this.

docs/guide/ethereum-provider.md
rekmarks: cf58561

docs/guide/ethereum-provider.md
Gudahtt: I can't see it anywhere else in state either. OK, maybe we just say it's specific to `eth_chainId` for now then, unless we discover otherwise.

```suggestion
- Ensure that chain IDs returned by `eth_chainId` are **not** 0-prefixed
```

docs/guide/ethereum-provider.md
rekmarks: ffa34ef

docs/guide/ethereum-provider.md
rekmarks: I want the world to know my pain, Mark!

docs/guide/ethereum-provider.md
rekmarks: No, we tell dapps to set it to `false` to silence its related warning, but it is set to `true` by us on initialization.

docs/guide/ethereum-provider.md
rekmarks: GitHub add " 😭" to your reactions you cowards.

docs/guide/ethereum-provider.md
Gudahtt: Huh. Nevermind then. I remembered this incorrectly.

docs/guide/ethereum-provider.md
rekmarks: 3d130b6

docs/guide/ethereum-provider.md
rekmarks: 3d130b6

docs/guide/ethereum-provider.md
rekmarks: 482ebb5

docs/guide/ethereum-provider.md
rekmarks: I'll remove any references to removal for now and we can resume this discussion for the migration guide.

docs/guide/ethereum-provider.md
rekmarks: 8f424f3

docs/guide/ethereum-provider.md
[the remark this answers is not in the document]
rekmarks: I believe I addressed this in: 3c76e9c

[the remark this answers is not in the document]
Gudahtt: Yes, I think so. I will create an issue in the `inpage-provider` repo about better representing connectivity.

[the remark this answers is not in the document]
Gudahtt: Here's the issue for improving connectivity: https://github.com/MetaMask/inpage-provider/issues/61

docs/guide/ethereum-provider.md
```diff
@@ -112,83 +182,394 @@ params: [
...
-See `ethereum.sendAsync`, directly below.
+This section documents our legacy provider API.
+MetaMask only supported this API before the provider API was standardized via [EIP-1193](https://eips.ethereum.org/EIPS/eip-1193) in 2020.
+Because of this, you may find web3 sites that use this API, or other providers that implement it.
 
-### ethereum.on(eventName, callback)
+## Legacy Properties
 
-The provider supports listening for some events:
+### ethereum.autoRefreshOnNetworkChange (DEPRECATED)
```
Gudahtt, line 399: I don't think this should be marked as deprecated yet either. It doesn't make sense to deprecate something we're still actively encouraging people to use. We'd be giving dapp developers contradictory instructions.
```suggestion
### ethereum.autoRefreshOnNetworkChange
```

rekmarks: 882eb8a

PR #134, opened 2020-06-26
v8 RPC API

rekmarks:
Updates the RPC API page for v8.

Assumes the experimental methods page is deleted.

### DIFF
--- a/docs/guide/rpc-api.md
+++ b/docs/guide/rpc-api.md
@@ -166,7 +166,7 @@ If the caller has no permissions, the array will be empty.
 
 ::: tip
 As an API consumer, you are unlikely to have to call this method yourself.
-Please see [`@metamask/onboarding`](https://github.com/MetaMask/metamask-onboarding) for more information.
+Please see the [Onboarding Library documentation](/onboarding-library.html) for more information.
 :::
 
 ```typescript
--- a/docs/guide/rpc-api.md
+++ b/docs/guide/rpc-api.md
@@ -1,5 +1,309 @@
-# The Ethereum RPC API
+# RPC API
 
-MetaMask uses the `ethereum.sendAsync()` (and soon, `ethereum.request()`) API to wrap an RPC API which is based on an interface exposed by all Ethereum clients, with some extra methods that are provided by MetaMask, as a key-holding signer. You can look up how to pass these methods to the `window.ethereum` object [here](./Ethereum-Provider).
+MetaMask uses the [`ethereum.request(args)` method](./ethereum-provider.html#ethereum-request-args) to wrap an RPC API.
 
-For the full API, please see [EIP 1474](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1474.md).
+The API is based on an interface exposed by all Ethereum clients, along with a growing number of methods that may or may not be supported by other wallets.
+
+::: tip
+All RPC method requests can return errors.
+Make sure to handle errors for every call to `ethereum.request(args)`.
+:::
+
+## Table of Contents
+
+[[toc]]
+
+## Ethereum JSON-RPC Methods
+
+For the Ethereum JSON-RPC API, please see [the Ethereum wiki](https://eth.wiki/json-rpc/API#json-rpc-methods).
+
+Important methods from this API include:
+
+- [`eth_accounts`](https://eth.wiki/json-rpc/API#eth_accounts)
+- [`eth_call`](https://eth.wiki/json-rpc/API#eth_call)
+- [`eth_getBalance`](https://eth.wiki/json-rpc/API#eth_getBalance)
+- [`eth_sendTransaction`](https://eth.wiki/json-rpc/API#eth_sendTransaction)
+- [`eth_sign`](https://eth.wiki/json-rpc/API#eth_sign)
+
+## Permissions
+
+These RPC methods are called in order to access the user's accounts.
+In the future, they will also be used to request access to other permissioned MetaMask functionality.
+
+MetaMask introduced Web3 Wallet Permissions via [EIP-2255](https://eips.ethereum.org/EIPS/eip-2255).
+In this permissions system, each RPC method is either _restricted_ or _open_.
+If a method is restricted, an external _domain_ (like a web3 site) must have the corresponding permission in order to call it.
+Open methods, meanwhile, do not require permissions to call, but may require confirmation by the user in order to succeed (e.g. `eth_sendTransaction`).
+
+Currently, the only permission is `eth_accounts`, which allows you to access the user's Ethereum address(es).
+More permissions will be added in the future.
+
+Under the hood, permissions are plain, JSON-compatible objects, with a number of fields that are mostly used internally by MetaMask.
+The following interface lists the fields that may be of interest to consumers:
+
+```typescript
+interface Web3WalletPermission {
+  // The name of the method corresponding to the permission
+  parentCapability: string;
+
+  // The date the permission was granted, in UNIX epoch time
+  date?: number;
+}
+```
+
+The permissions system is implemented in the [`rpc-cap` package](https://github.com/MetaMask/rpc-cap).
+If you're interested in learning more about the theory behind this _capability_-inspired permissions system, we encourage you to take a look at [EIP-2255](https://eips.ethereum.org/EIPS/eip-2255).
+
+### eth_requestAccounts
+
+::: tip EIP-1102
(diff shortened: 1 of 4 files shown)


docs/guide/rpc-api.md
```diff
@@ -1,5 +1,309 @@
...
+### eth_decrypt
+
+```typescript
+// these two types are simply hexadecimal strings
+type EthereumAddress: string;
+type EncryptedMessage: string;
+// the decrypted message is the originally encrypted string
+type DecryptedMessage: string;
+
+ethereum.request({
```
brad-decker, line 285: I get that the goal of this is to inform the user of the types, but this wouldn't compile because its a mixture of a function invocation with a type signature for that function invocation. You'd need to make ethereum.request a parameterized function eg:

```ts
function request<ParamsType, ReturnType>({ method: string, params: ParamsType }): Promise<ReturnType> {
  // implementation
}

request<[EncryptedMessage, EthereumAddress], DecryptedMessage>({ method: 'eth_decrypt', params: ['0xasdf1a7=', '0xa00a..324r3']
}).then(decryptedMessage => { // do something with decryption });
```

If we have an explainer somewhere that the examples are just for showing types and not real code, then we can just leave it as is maybe?

rekmarks: Hmm, yeah, the goal is mostly to be informative, and I think for most devs, the parameterized function obscures more than it explains, even though it compiles 🤔 

I'm inclined to leave a notice, depending on what we all think. It is a bit ugly to have it listed as TS that's just... syntactically invalid.

brad-decker: alternatively...

```ts
// Create an intermediary typesafe function
function ethDecrypt(params: [EncryptedMessage, EthereumAddress]): Promise<DecryptedMessage> {
  return ethereum.request({ method: 'eth_decrypt', params })
}
```

might be cleaner in regards to not relying on Algebraic data types, compiles, etc.

rekmarks: Hm, or, alternatively alternatively, I suppose we do what most JSON-RPC APIs I’ve seen do, e.g. https://kodi.wiki/view/JSON-RPC_API/v10

It’s the same for the Ethereum JSON-RPC API listed on eth.wiki.

docs/guide/rpc-api.md
```diff
@@ -1,5 +1,309 @@
-# The Ethereum RPC API
+# RPC API
 
-MetaMask uses the `ethereum.sendAsync()` (and soon, `ethereum.request()`) API to wrap an RPC API which is based on an interface exposed by all Ethereum clients, with some extra methods that are provided by MetaMask, as a key-holding signer. You can look up how to pass these methods to the `window.ethereum` object [here](./Ethereum-Provider).
+MetaMask uses the [`ethereum.request(args)` method](./ethereum-provider.html#ethereum-request-args) to wrap an RPC API.
 
-For the full API, please see [EIP 1474](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1474.md).
```
Gudahtt, line 5: Now that the title of this page has changed, it no longer seems appropriate to call this the "full" API. The Ethereum wiki link below documents a wider variety of RPC methods than this EIP does, and even that doesn't show them all (e.g. `eth_subscribe`)

docs/guide/rpc-api.md
```diff
@@ -1,5 +1,309 @@
...
+Returns a Promise that resolves to an array of `Web3WalletPermission` objects.
+If the caller has no permissions, the array will be empty.
+
+## Other RPC Methods
+
+### wallet_registerOnboarding
+
+::: tip
+As an API consumer, you are unlikely to have to call this method yourself.
+Please see [`@metamask/onboarding`](https://github.com/MetaMask/metamask-onboarding) for more information.
```
Gudahtt, line 169: We should link the docs page about onboarding here instead. The repo README isn't really intended for dapp devs.
```suggestion
Please see the [Onboarding Library documentation](/onboarding-library.html) for more information.
```

docs/guide/rpc-api.md
```diff
@@ -1,5 +1,309 @@
...
+:::
+
+```typescript
+ethereum.request({ method: 'wallet_registerOnboarding' }): Promise<true>;
+```
+
+Requests that MetaMask completes its onboarding flow for the user.
```
Gudahtt, line 176: Whoops; this is not correct.
```suggestion
This method is used to register the requesting site with MetaMask as the initiator of onboarding. In other words, you can use this method to inform MetaMask that you were the one that suggested installing MetaMask. This lets MetaMask redirect the user back to your site after onboarding as completed.
```

docs/guide/rpc-api.md
```diff
@@ -1,5 +1,309 @@
...
+Returns a Promise that resolves to `true`, or rejects if there's an error.
+
+This method is intended to be called at the end of the MetaMask onboarding flow.
```
Gudahtt, line 179:
```suggestion
This method is intended to be called after MetaMask has been installed, but before the MetaMask onboarding has completed.
```

docs/guide/rpc-api.md
```diff
@@ -1,5 +1,309 @@
...
+The onboarding flow is initiated by our [`@metamask/onboarding`](https://github.com/MetaMask/metamask-onboarding) package, which calls `wallet_registerOnboarding` upon redirecting the user back to the original site.
```
Gudahtt, line 180: This sentence seems incorrect. I'm not sure we need to replace it with anything.
```suggestion
```

docs/guide/rpc-api.md
```diff
@@ -1,5 +1,309 @@
...
+::: tip EIP-747
+This method is specified by [EIP-747](https://eips.ethereum.org/EIPS/eip-747).
+:::
+
+```typescript
+interface WatchAssetParams {
+  type: 'ERC20', // In the future, other standards will be supported
+  options: {
+    address: string, // The address of the token contract
+    'symbol': string, // A ticker symbol or shorthand, up to 5 characters
```
Gudahtt, line 194: Why quote just this property :thinking: 
```suggestion
    symbol: string, // A ticker symbol or shorthand, up to 5 characters
```

rekmarks: Because `symbol` is a reserved keyword, and the syntax highlighting picks it up.

Not sure if it'd be better to quote everything?

Gudahtt: Ah, no, this seems fine. I totally forgot it was a reserved keyword.

docs/guide/rpc-api.md
Gudahtt: Maybe we can link nothing on this line, and link both below? 🤔 Not sure.

docs/guide/rpc-api.md
rekmarks: 2862988

docs/guide/rpc-api.md
```diff
@@ -1,5 +1,309 @@
...
+ethereum.request({ method: 'eth_requestAccounts' }): Promise<string[]>
+```
+
+Requests that the user provides an Ethereum address to be identified by.
+Returns a Promise that resolves to an array of a single Ethereum address string.
+If the user denies the request, the Promise will reject with a `4001` error.
+
+The request causes a MetaMask popup to appear.
+You should only request the user's accounts in response to user action, such as a button click.
+Otherwise, you will popup-spam the user.
```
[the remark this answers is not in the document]
rekmarks: 2862988

docs/guide/rpc-api.md
```diff
@@ -1,5 +1,309 @@
...
+- [`eth_accounts`](https://eth.wiki/json-rpc/API#eth_accounts)
+- [`eth_call`](https://eth.wiki/json-rpc/API#eth_call)
+- [`eth_getBalance`](https://eth.wiki/json-rpc/API#eth_getBalance)
+- [`eth_sendTransaction`](https://eth.wiki/json-rpc/API#eth_sendTransaction)
+- [`eth_sign`](https://eth.wiki/json-rpc/API#eth_sign)
+
+## Permissions
+
+These RPC methods are called in order to access the user's accounts.
+In the future, they will also be used to request access to other permissioned MetaMask functionality.
```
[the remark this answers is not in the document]
rekmarks: 2862988

docs/guide/rpc-api.md
rekmarks: @Gudahtt @brad-decker this is the last thing we need to check off for this PR. See: 3dc33db

The addition of `.prettierrc.js` was necessary to create an exception for a particular snippet that used a reserved TypeScript keyword (`symbol`) in an interface.

PR #141, opened 2020-07-02
v8 - Add Rollout Notices

rekmarks:
Add rollout notices to the Provider and RPC API pages of the docs so we can get this into production!


rekmarks:
@Gudahtt I threw in a "hopefully" for that reason, making it no more specific than this Tweet, at the end of the day: https://twitter.com/metamask_io/status/1278698985142628353?s=20

PR #146, opened 2020-07-09
Fix `metamask_watchAsset` example

Gudahtt:
The example passed an array as the `params` property, which does not work. `params` must be passed as an object.


Gudahtt:
This was originally reported here: https://github.com/MetaMask/metamask-extension/issues/8945

PR #147, opened 2020-07-09
Fix `metamask_watchAsset` syntax error

Gudahtt:
The `wasAdded` variable was referenced by the wrong name (`added`) and from the wrong scope. It now checks the result correctly.

### DIFF
--- a/docs/guide/registering-your-token.md
+++ b/docs/guide/registering-your-token.md
@@ -23,8 +23,8 @@ const tokenSymbol = 'TUT';
 const tokenDecimals = 18;
 const tokenImage = 'http://placekitten.com/200/300';
 
-// wasAdded is a boolean. Like any RPC method, an error may be thrown.
 try {
+  // wasAdded is a boolean. Like any RPC method, an error may be thrown.
   const wasAdded = await ethereum.request({
     method: 'metamask_watchAsset',
     params: {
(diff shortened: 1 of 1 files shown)


docs/guide/registering-your-token.md
```diff
@@ -23,8 +23,8 @@ const tokenSymbol = 'TUT';
 const tokenDecimals = 18;
 const tokenImage = 'http://placekitten.com/200/300';
 
-// wasAdded is a boolean. Like any RPC method, an error may be thrown.
 try {
+  // wasAdded is a boolean. Like any RPC method, an error may be thrown.
   const wasAdded = await ethereum.request({
     method: 'metamask_watchAsset',
```
rekmarks, line 29:
```suggestion
    method: 'wallet_watchAsset',	
```

Gudahtt: Hmm, I guess so 🤔 

The current EIP doesn't really clarify which one we should use. It refers to a _function_ called `wallet_watchAsset` (and also called `web3.wallet.watchAsset`), but doesn't reference a _method_ with that name. The only method shown is `metamask_watchAsset`, in the example GIF.

rekmarks: Yeah, I think the intention of @danfinlay and @estebanmino is for it to be `wallet_watchAsset`. This is also in line with EIP-2255. In addition, if we're hoping for it to be adopted by others, we should probably not prefix the method with our brand name.

Our implementation still accepts `metamask_watchAsset` because we _have_ put that out there.

rekmarks: We might be able to automerge an update to that EIP, if we want to clean it up.

PR #163, opened 2020-08-14
Update provider documentation with deprecation dates

rekmarks:
We've announced dates for breaking provider API changes and the web3 removal. This updates the docs in light of this.

### DIFF
--- a/docs/guide/provider-migration.md
+++ b/docs/guide/provider-migration.md
@@ -3,7 +3,7 @@
 ::: danger Deadline
 All of these changes are shipping on **November 16, 2020**.
 
-If your web3 site relies on any of these features, it will break when they are removed unless you migrate.
+If your web3 site relies on any of these features, it may break when these changes are made unless you migrate.
 :::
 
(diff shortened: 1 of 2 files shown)


docs/guide/provider-migration.md
```diff
@@ -1,6 +1,12 @@
 # Provider Migration Guide
 
-As noted in the [Ethereum Provider documentation](./ethereum-provider.html#upcoming-breaking-changes), we are going to make breaking changes to our provider API and remove our injected `window.web3`.
+::: danger Deadline
+All of these changes are shipping on **November 16, 2020**.
+
+If your web3 site relies on any of these features, it will break when they are removed unless you migrate.
```
Gudahtt, line 6: Slight rewording, since at least one of these changes (the `eth_chainId` return value) is a change and not a removal, and _might_ not break anything for a site that used the feature.
```suggestion
If your web3 site relies on any of these features, it may break when these changes are made unless you migrate.
```

PR #168, opened 2020-08-28
Add Algolia DocSearch secrets to CI build script

rekmarks:
This adds [Algolia DocSearch](https://docsearch.algolia.com/) as our search provider, enabling full-text search for our docs. The default VuePress search provider is an-inline JavaScript implementation that only searches Markdown headers. Algolia is used by a ton of popular JavaScript projects, including Babel, TypeScript, React, and Bootstrap.

- Add `dotenv` dev dependency for Algolia DocSearch secrets
- Create `build.js` for local builds
  - Uses `dotenv`
- Create `ci-build.sh` for CI builds
  - Uses `build.js`
  - Sets environment variables for build script via hidden CI environment variables
- Update file deletion command in deploy script
  - To ensure that `.env-template` is deleted
- Add instructions for creating `.env` file in readme

PR #170, opened 2020-08-31
Fix deploy script

rekmarks:
#168 introduced `find` in the deploy script in order to delete files and directories that shouldn't be deployed. The `-delete` option errors when attempting to delete non-empty directories, so that didn't work. In addition, we were deleting some files that should be preserved.

This PR uses `rm -rf` in combination with `grep` to delete all files and directories except `.git` and the ones not located in `docs/dist`.

[Example successful CI run.](https://app.circleci.com/pipelines/github/MetaMask/metamask-docs/706/workflows/2d17f5b0-9bc4-42c2-9b1c-1fb862a81529/jobs/1246)


whymarrh:
[Parsing `ls` isn't ideal](https://mywiki.wooledge.org/ParsingLs), could we add `-type f` to stick to deleting files—would something like that work?

rekmarks:
I can continue fighting with `find`, but how likely are we to add files to the root of this repository with newlines in their names? I agree that parsing `ls` is bad in the general case, but this isn't the general case.

rekmarks:
We also want to delete files _and_ directories; the VuePress build artifacts have different names for each build, so we're actually publishing... _a lot_ of cruft: https://github.com/MetaMask/metamask-docs/tree/gh-pages/assets/js

rekmarks:
@whymarrh I looked into it some more, and if we want to handle the unlikely case of newlines in filenames, someone else is going to have to do it.

PR #178, opened 2020-10-07
Skip Algolia env checks for non-main builds

whymarrh:
This PR skips the check for the Algolia environment variables on main builds (e.g. build from PRs). Once the changes land on the main branch they will be built with the correct keys.


whymarrh:
Whoops, meant to test this coming from a fork 🤦🏽‍♂️

PR #197, opened 2020-11-10
Raise web3 deprecation to top of provider migration guide

danfinlay:
I thought it was weird that we highlight how to migrate off multiple
undocumented, unsupported APIs before we highlight the real star: The
web3 API itself.

I suspect the vast majority of people visiting this page only need to
read the section on web3.js, and so I think it should be the top topic
addressed.


rekmarks (approved):
LGTM!

PR #198, opened 2020-11-18
Update breaking changes comms

rekmarks:
Updates communications about breaking changes to reflect that we are _in the process of_ shipping them, and that they may land any day.


BboyAkers (approved):
LGTM

PR #199, opened 2020-12-01
Add signTypedData_v4 in history section

frangio:
Is this right? I thought it was weird that the v4 method has its own section but isn't mentioned in the list.


BboyAkers (approved):
Nice!! Thanks for the catch! 🙂

PR #219, opened 2021-01-27
use correct example repo

BitHighlander:
The BboyAkers for repo appears out of date. it wont connect and giving the error. ```We noticed that the current website tried to use the removed window.web3 API. If the site appears to be broken, please click here for more information.``` https://metamask.zendesk.com/hc/en-us/articles/360053147012

I believe this repo is what was intended


BboyAkers:
It the correct repo, I'm currently updating the tutorial 🙂. My apologies for the delay!

PR #220, opened 2021-01-27
Fix broken links in README

oskarth:
Currently at https://docs.metamask.io/guide/ links 404 because it is linking to ./guide/guide/foobar.html instead of ./guide/foobar.html. The README is at the same top level as the other guide docs.


rekmarks (approved):
Tack så mycket, chefen!

Eller bör jag säga, 太谢谢你了来自瑞典和会说中文的朋友。

PR #221, opened 2021-01-28
modify the wrong rpc name:eth_requestAccounts

zctocm:
modify the wrong rpc name:eth_requestAccounts to wallet_requestPermissions on rpc-api doc


Gudahtt (approved):
Great catch, thanks!

PR #235, opened 2021-03-20
Update RPC method anchor links to eth.wiki

johnrees:
I haven't tested this in other browsers but in Brave, https://eth.wiki/json-rpc/API#eth_sendTransaction does not jump to the correct section on the eth.wiki page, whereas https://eth.wiki/json-rpc/API#eth_sendtransaction does. Therefore I updated some of the links to lowercase to fix that problem.


BboyAkers (approved):
I love it! Thank you for catching this! 😃

PR #254, opened 2021-06-06
Fix syntax error in example code

marcinm2h:
Line 402 - unnecessary semicolon - syntax error. Also added extra new line after "ethereum" for style consistency in this file.


ryanml (approved):
thanks for this!

PR #269, opened 2021-08-13
Fix twitter URL to the current account being used

flada-auxv:
I have updated the twitter URL as https://twitter.com/MetaMask seems to be in use now. I don't know why, but old one (https://twitter.com/metamask_io) seems to be suspended.


ryanml (approved):
Awesome, thank you for this! :)

PR #275, opened 2021-09-03
Add docs for removeListener in ethereum-provider.md

pmzi:
Add documentation for `ethereum.removeListener` which is highly useful especially when developing with React.
I know `removeListener` can be derived from [NodeJS events](https://nodejs.org/api/events.html#events_emitter_removelistener_eventname_listener); however, because of its importance, it's better to be mentioned inside docs explicitly.

Thanks!


Gudahtt (approved):
Good idea, thanks! LGTM

Issue #265, opened 2021-07-21
Issue with eth_signTypedData_v4 signature.


Vivek205:
I'm creating an eth_signTypedData_v4 signature using Metamask. 
Whenever I change the domain name in the data, my signature is getting affected. 
But as per the [Metamask documentation](https://docs.metamask.io/guide/signing-data.html#sign-typed-data-v4), the `domain name` is only for UX purposes.
So changing `domain name` should not change the signature. 

### Case 1. - Signature with domain name = "Uniswap V2"
Data:
```
{"types":{"EIP712Domain":[{"name":"name","type":"string"},{"name":"version","type":"string"},{"name":"chainId","type":"uint256"},{"name":"verifyingContract","type":"address"}],"Permit":[{"name":"owner","type":"address"},{"name":"spender","type":"address"},{"name":"value","type":"uint256"},{"name":"nonce","type":"uint256"},{"name":"deadline","type":"uint256"}]},"domain":{"name":"Uniswap V2","version":"1","chainId":3,"verifyingContract":"0x4C78b6566864Ae6304c2c2A4c43B74DAFaAc167E"},"primaryType":"Permit","message":{"owner":"0xabd2cCb3828b4428bBde6C2031A865b0fb272a5A","spender":"0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D","value":"18677329819741772","nonce":1,"deadline":1626847711}}
```
Signature:
```
0x3b8115c48129883d1308aa6f09083e0ae9cc748d5af2e5ea4ae51de597738b26184a7cf9eb02ee1a8d0a045f7461312783681fcbd033bf6d7247caca76bb3aac1b
```

### Case 2. - Signature with domain name = "Different Name"

Data:
```
[...]

PR #277, opened 2021-09-08
Added comment for name in SignTypedData V4

BboyAkers:


BboyAkers:
@Gudahtt Please let me know if we need to document this field more clearly 🙂

Gudahtt:
I think the primary confusion in #265 was about whether the name was incorporated into the signature itself. The user thought that "primarily for UX purposes" meant that it wasn't included. I don't think repeating that in the example is likely to prevent future readers from getting that same impression. It's not correct that it's "needed for the signature" either. It's optional, but if provided it will be incorporated into the signature.

I'm not sure how best to explain that all of these parameters are included in the signature :thinking: Maybe with a sentence above the list of parameters? It applies to all of them really.

BboyAkers:
> I think the primary confusion in #265 was about whether the name was incorporated into the signature itself. The user thought that "primarily for UX purposes" meant that it wasn't included. I don't think repeating that in the example is likely to prevent future readers from getting that same impression. It's not correct that it's "needed for the signature" either. It's optional, but if provided it will be incorporated into the signature.
> 
> I'm not sure how best to explain that all of these parameters are included in the signature 🤔 Maybe with a sentence above the list of parameters? It applies to all of them really.

I'm not sure either on how to explain them. I'll reach back out to Remco for advice since he's an individual who submitted this EIP 🙂.

BboyAkers:
I've messaged Remco and still haven't heard back. Anyone have an alternate way of contacting him? @danfinlay @rekmarks @brad-decker

Lucid36 (approved):

alexandratran (approved):
Updated to the new docs and added a simple comment saying that all parameters affect the signature. LGTM.

PR #281, opened 2021-09-10
Typo fix

daniel-farina:
Fixed typo of using MetaMaskOnboarding instead of MetamaskOnboarding. (Note the capital M)
Reference: https://github.com/BboyAkers/simple-dapp-tutorial/issues/26

### DIFF
--- a/docs/guide/create-dapp.md
+++ b/docs/guide/create-dapp.md
@@ -188,7 +188,7 @@ Above your `MetaMaskClientCheck` function write/insert this code.
 
 ```javascript
 //We create a new MetaMask onboarding object to use in our app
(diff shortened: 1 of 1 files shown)


Gudahtt (commented):
Thanks for the contribution! Sorry about this mixup but this is in fact the correct capitalization for this global. It was called `MetamaskOnboarding` in an older pre-v1 version of the onboarding library, but it's called `MetaMaskOnboarding` from v1 onwards.

I think the better approach would be to merge this PR instead: https://github.com/BboyAkers/simple-dapp-tutorial/pull/31/files

Gudahtt:
cc @BboyAkers, could you take a look at that PR when you get the chance? :pray: I'll close this PR in the meantime.

PR #282, opened 2021-09-18
Fixing a very cumbersome sentence in the intro.

cliffhall:
> Studying the history of these methods has some lessons in it for the emergent lessons of decentralized standards emergence.

becomes

> Studying the history of these methods yields some guiding lessons for the emergence of decentralized standards.


BboyAkers (approved):
LGTM 🙂

PR #293, opened 2021-11-12
fix:typo in signing-data

bestony:
fix typo in documents


Gudahtt (approved):
Thanks, good catch! LGTM

Issue #332, opened 2022-01-12
Update Create a Simple Dapp


httpJunkie:
The "Create a Simple Dapp" tutorial on the MetaMask docs site has a few instructions that we can add to ensure better readability and a better understanding of where the code samples in the tutorial should be placed.

I had a few issues but worked through them looking at the `/finished/index.js` file. I would like to update the tutorial to ensure they don't have to do the same.

- The following two code blocks need some simple explanation as to where to paste them. (**edit:** only one)

<img width="731" alt="Screen Shot 2022-01-11 at 7 06 05 PM" src="https://user-images.githubusercontent.com/4899480/149056968-ce9df8c7-1cc5-452f-a401-de93b8413a9f.png">

Not a huge issue, but will be consistent with the rest of the tutorial where we held their hand a little more.

- I would like to make changes to the overall text and readability of the tutorial
- Add a final code block incase someone has had any issues that they do not have to check the `/finished` directory and track down a file with another name `index.js` in order to troubleshoot where they went wrong.

Overall, the tutorial was easy enough to get through. This will be a good opportunity for a first PR and to start collaborating with @BboyAkers.

BboyAkers:
Merged awhile ago! Going to close 🙂

PR #333, opened 2022-01-12
Update Create Simple dApp Tutorial

httpJunkie:
This PR is in regards to the MetaMask GitHub #332 

Contains no code changes to the tutorial but does add a new code block at the end so the user can see the final state of the code progress worked through in this first part of the tutorial.

It also contains grammatical, typo, clarity and brevity fixes without losing the intent of the original author.

The overall idea is to make the tutorial more clear and readable.


BboyAkers (approved):
LGTM

PR #340, opened 2022-01-16
Amend readme example for encrypting with `eth-sig-utils`

aa-neg:
The current example utilizing  [`eth-sig-util`](https://github.com/MetaMask/eth-sig-util) doesn't use name parameters and currently has the publicKey in the [wrong argument position](https://github.com/MetaMask/eth-sig-util/blob/main/src/encryption.ts#L22). 

This PR converts to named parameters to avoid this issue.

### DIFF
--- a/docs/guide/rpc-api.md
+++ b/docs/guide/rpc-api.md
@@ -277,7 +277,7 @@ const encryptedMessage = ethUtil.bufferToHex(
     JSON.stringify(
       sigUtil.encrypt(
         {
-          pubKey: encryptionPublicKey,
+          publicKey: encryptionPublicKey,
           data: 'hello world!,
           version: 'x25519-xsalsa20-poly1305',
         }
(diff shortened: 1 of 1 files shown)


docs/guide/rpc-api.md
```diff
@@ -277,9 +277,11 @@ const encryptedMessage = ethUtil.bufferToHex(
   Buffer.from(
     JSON.stringify(
       sigUtil.encrypt(
-        encryptionPublicKey,
-        { data: 'Hello world!' },
-        'x25519-xsalsa20-poly1305'
+        {
+          pubKey: encryptionPublicKey,
```
Gudahtt, line 281: This parameter name should be `publicKey`:
```suggestion
          publicKey: encryptionPublicKey,
```

Gudahtt (changes requested):
Thanks! I hadn't realized `eth-sig-util` was used directly here in the docs, this should have been updated with the latest release.

There is still one mistake but this looks good

Gudahtt (approved):
LGTM!

PR #345, opened 2022-01-17
Fix lint script

rekmarks:
Ensure that `yarn lint` exits with a non-zero code if linting fails. Just running `prettier` without the `--check` flag is pretty much useless. This will prevent unlinted files from being merged to `main`.

The `lint:fix` script still works because the `--write` flag appears to effectively override the `--check` behavior.


Gudahtt (approved):
LGTM!

BboyAkers (approved):

PR #347, opened 2022-01-17
Use backticks for RPC method subheaders

Gudahtt:
The subheaders for each RPC method on the RPC API page now use backticks, to better communicate that these are exact names and not just titles.


rekmarks (approved):
LGTM!

shanejonas (approved):

PR #348, opened 2022-01-17
Update permission system documentation

Gudahtt:
The RPC API documentation about the permission system has been updated to reflect the current terminology we use, and to reflect recent changes.

The term "open methods" has been dropped in favor of "unrestricted methods", which is more in-line with "restricted methods" and is what we have been calling them in practice.

The wording was updated to no longer imply that a _domain_ is the permission subject. Now it just says "caller" instead, which is just as clear but leaves open the possibility that the caller is not a website. This was done in preparation for Snaps, since snaps can also call RPC methods.

The definition for "Unrestricted methods" was updated to clarify that sometimes they do depend upon permissions indirectly. The unrestricted methods just don't have permissions that directly correspond with the method name.

The note on the implementation was removed, since the permission system no longer uses `rpc-cap`. I removed it rather than linking the new implementation because it's not really in a state where it would be easy to understand or re-use.

The "Other RPC Methods" header has been renamed to "Unrestricted Methods". The old title was silly; this section was alongside the header "Restricted Methods" so it always meant to refer to unrestricted methods.

### DIFF
--- a/docs/guide/rpc-api.md
+++ b/docs/guide/rpc-api.md
@@ -34,7 +34,7 @@ Important methods from this API include:
 MetaMask introduced Web3 Wallet Permissions via [EIP-2255](https://eips.ethereum.org/EIPS/eip-2255).
 In this permissions system, each RPC method is either _restricted_ or _unrestricted_.
 If a method is restricted, the caller must have the corresponding permission in order to call it.
-Unrestricted methods, meanwhile, do not require permissions to call, but may require confirmation by the user in order to succeed (e.g. `wallet_addEthereumChain`).
+Unrestricted methods, meanwhile, have no corresponding permission. Some of them still rely upon permissions to succeed though (e.g. the signing methods require that you have the `eth_accounts` permission for the signer account), and some require confirmation by the user (e.g. `wallet_addEthereumChain`).
 
 Currently, the only permission is `eth_accounts`, which allows you to access the user's Ethereum address(es).
 More permissions will be added in the future.
@@ -177,6 +177,8 @@ function requestPermissions() {
 }
 ```
 
+## Unrestricted Methods
+
 ### `eth_decrypt`
 
 ::: tip Platform Availability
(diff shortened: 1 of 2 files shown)


docs/guide/rpc-api.md
```diff
@@ -178,8 +177,6 @@ function requestPermissions() {
 }
 ```
 
-## Other RPC Methods
-
```
rekmarks, line 182: I believe that the unrestricted methods section should start here.

rekmarks (changes requested):

docs/guide/rpc-api.md
Gudahtt: See the note in the PR description:

>The decryption methods were pulled into the "restricted methods" section because they are restricted methods.

rekmarks: Heretofore I've defined "restricted methods" as a method that has a permission, not a method that has a confirmation. The decryption methods are in our unrestricted methods constants that are passed to the permission controller.

rekmarks: Ref: https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/controllers/permissions/specifications.js/#L205

rekmarks: Or you mean to say that it's de facto restricted because it relies on `eth_accounts`?

Gudahtt: Yes exactly. It's restricted by the eth accounts permission. It returns an error without it.

Gudahtt: We discussed this on Slack and decided to leave them as "Unrestricted" for now, but to ensure that the definition of "Unrestricted" didn't imply that no permission was needed. Just that there was no permission with a name matching the RPC method name. Also we should add a warning explaining that the permission is needed, but that can be in a separate PR. I will update this PR.

Gudahtt: Addressed in https://github.com/MetaMask/metamask-docs/pull/348/commits/c42ac389bef377113968295b5537f6118df0300a, and the PR description has been updated.

rekmarks (approved):
LGTM!

PR #368, opened 2022-01-21
Update create-dapp.md

rarcif:
preferred package in root makes use of yarn. npm fails.


Gudahtt (approved):
LGTM!

It looks like some other details of these instructions need to be updated to match changes in the sample dapp as well (e.g. the list of files to expect). But this is a step forward.

PR #370, opened 2022-01-26
Update getting-started.md

madforce:
Fleshing out the instructions for connecting to test networks and using Ganache. The instructions looked to be out of date refering to RPC etc (account is in top right, so made it clear with some screenshots) which is in the advanced section. Rather than refering to displaying of preconfigured test networks or just using the "Add Network" function in the network list.


BboyAkers:
Thanks for this PR! @Gudahtt can you take a look as well? 🙂

docs/guide/getting-started.md
```diff
@@ -26,11 +26,51 @@ if (typeof window.ethereum !== 'undefined') {
...
 
+### Connecting to a test network.
+
+**Show test networks**<br />
+![image](https://user-images.githubusercontent.com/1038445/150907727-77fe4899-5238-4df6-8fa4-e6ae522e7d02.png)<br />
+
+![image](https://user-images.githubusercontent.com/1038445/150907769-c7ea63a8-6bd2-4f80-a376-a70243bb14b7.png)<br />
+
+**Connecting to Ropsten Test Network**<br />
+Use the network dropdown menu (which would normally be displaying Ethereum Mainnet like the image above) and select Ropsten Test Network<br />
```
mcmire, line 37: Can you add a colon at the end of this line? Also, again, you shouldn't need `<br />`.

docs/guide/getting-started.md
```diff
@@ -26,11 +26,51 @@ if (typeof window.ethereum !== 'undefined') {
 
 You can review the full API for the `window.ethereum` object [here](./ethereum-provider.html).
 
+### Connecting to a test network.
```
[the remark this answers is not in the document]
mcmire: It looks like this section is more about just connecting to a test network: it shows you how to set up a test network for use in development. So maybe this should be "Setting Up a Test Network"?

docs/guide/getting-started.md
```diff
@@ -26,11 +26,51 @@ if (typeof window.ethereum !== 'undefined') {
...
+
+**Show test networks**<br />
+![image](https://user-images.githubusercontent.com/1038445/150907727-77fe4899-5238-4df6-8fa4-e6ae522e7d02.png)<br />
+
+![image](https://user-images.githubusercontent.com/1038445/150907769-c7ea63a8-6bd2-4f80-a376-a70243bb14b7.png)<br />
+
+**Connecting to Ropsten Test Network**<br />
```
mcmire, line 36: If we put this here, then it will sound like the only test network that MetaMask officially supports for development purposes is Ropsten. There is nothing wrong with this if this is true — and I think it is — but maybe it's worth adding some words about this after this header in case people are curious? Otherwise I think the instructions would seem very specific for a reason that might not be obvious.

docs/guide/getting-started.md
```diff
@@ -26,11 +26,51 @@ if (typeof window.ethereum !== 'undefined') {
...
+Use the network dropdown menu (which would normally be displaying Ethereum Mainnet like the image above) and select Ropsten Test Network<br />
+![image](https://user-images.githubusercontent.com/1038445/150908059-f86cd2eb-6d59-43e8-8fec-bacce425ebe0.png)<br />
+
+I recommend creating a test account just for the Ropsten Test Network. (Just give it a name)<br />
```
mcmire, line 40: If this is important, then perhaps this should be the first step in this section?

docs/guide/getting-started.md
```diff
@@ -26,11 +26,51 @@ if (typeof window.ethereum !== 'undefined') {
...
+
+![image](https://user-images.githubusercontent.com/1038445/150908294-22099c19-4b22-4587-939b-3f9f882b09e8.png)
+
+**Request 1 ETH Over the test network**<br />
```
mcmire, line 44: In case we update the faucet at some point, maybe we want to be less explicit here? For example:

``` suggestion
#### 3. Obtaining Test Ether
```

docs/guide/getting-started.md
```diff
@@ -26,11 +26,51 @@ if (typeof window.ethereum !== 'undefined') {
...
+Navigate to https://faucet.metamask.io/
+Ensure you have the test account and test "Ropsten Test Network Selected"
+Click "reqest 1 ether from faucet"
+![image](https://user-images.githubusercontent.com/1038445/150909358-80beed6d-bebc-4602-90fb-183b15279228.png)
+
+A request will be sent to Metamask Client for you to sign. Click Next **WAIT Patiently for the transaction to be approved once Metamask has connected (may take a couple minutes)**
```
mcmire, line 50: A few things:

* Metamask -> "MetaMask"
* You can drop "Client" here — I think it's obvious we're talking about the client
* I think using bold here is not necessary, and you might be missing a period

Putting this together, what do you think about:

``` suggestion
A request will be sent to MetaMask for you to sign. Click Next, and wait for the transaction to be approved (this may take a couple minutes):
```

docs/guide/getting-started.md
```diff
@@ -26,11 +26,51 @@ if (typeof window.ethereum !== 'undefined') {
...
+![image](https://user-images.githubusercontent.com/1038445/150909613-6cb6d54e-3d43-4505-b8d6-7fa6b225e355.png)
+
+![image](https://user-images.githubusercontent.com/1038445/150909688-211f3398-e737-42db-bee2-e62d5e2ef5a8.png)
+
+![image](https://user-images.githubusercontent.com/1038445/150910232-703e1c26-4bdb-41aa-a174-7c10a9f3a6f8.png)
```
mcmire, line 55: I don't think the "User" section needs to be highlighted here. This shows your current balance but also allows you to donate to the faucet. Maybe it's worth showing a screenshot of the pending or confirmed transaction in the transaction list in MetaMask instead?

docs/guide/getting-started.md
```diff
@@ -26,11 +26,51 @@ if (typeof window.ethereum !== 'undefined') {
...
+
+![image](https://user-images.githubusercontent.com/1038445/150910318-b6cd9eb6-56e8-46b0-9664-82a072d3b0a5.png)
+
 ### Running a Test Network
 
-In the top right menu of MetaMask, select the network that you are currently connected to. Among several popular defaults, you'll find `Custom RPC` and `Localhost 8545`. These are both useful for connecting to a test blockchain, like [ganache](https://www.trufflesuite.com/ganache). You can quickly install and start Ganache if you have `npm` installed with `npm i -g ganache-cli && ganache-cli`.
+If you would like to run your own Network on you local machine. [Ganache](https://www.trufflesuite.com/ganache) has some great features for starting your application with different states.
```
mcmire, line 61: Typo:

``` suggestion
If you would like to run your own network on your local machine, [Ganache](https://www.trufflesuite.com/ganache) has some great features for starting your application with different states.
```

docs/guide/getting-started.md
```diff
@@ -26,11 +26,51 @@ if (typeof window.ethereum !== 'undefined') {
...
+
+You can quickly install and start Ganache if you have `npm` installed with `npm i -g ganache-cli && ganache-cli`.
+Once Ganache is started connect to network `Localhost 8545` at the bottom of the network list (if you don't see `Localhost 8545` in your network list follow the instructions above [`Connecting to a test network`](#Connecting) to `show` test networks).
```
mcmire, line 64: Just some cleanup:

``` suggestion
Once Ganache is started, connect to the network `Localhost 8545`, which you can find at the bottom of the network list (if you don't see this, follow the instructions in [`Connecting to a test network`](#Connecting) to display test networks).
```

docs/guide/getting-started.md
```diff
@@ -26,11 +26,51 @@ if (typeof window.ethereum !== 'undefined') {
...
+![image](https://user-images.githubusercontent.com/1038445/150908059-f86cd2eb-6d59-43e8-8fec-bacce425ebe0.png)<br />
+
+I recommend creating a test account just for the Ropsten Test Network. (Just give it a name)<br />
+
+![image](https://user-images.githubusercontent.com/1038445/150908294-22099c19-4b22-4587-939b-3f9f882b09e8.png)
+
+**Request 1 ETH Over the test network**<br />
+Navigate to https://faucet.metamask.io/
+Ensure you have the test account and test "Ropsten Test Network Selected"
+Click "reqest 1 ether from faucet"
```
mcmire, line 47: The line breaks here may not translate to the docs. Also, "ensure you have the test account" doesn't quite make sense. What do you think about saying:

``` suggestion
Navigate to [MetaMask's test faucet](https://faucet.metamask.io/). Then, click on "request 1 ether from faucet" and select the test account you created earlier:
```

docs/guide/getting-started.md
```diff
@@ -26,11 +26,51 @@ if (typeof window.ethereum !== 'undefined') {
...
+You can now 
+![image](https://user-images.githubusercontent.com/1038445/150916135-c9d019ad-86f3-474d-a0ef-d53dabe799f9.png)
+
+If your application starts with the `-m` (bip39 mnemonic phrase for generating a PRNG seed) flag, you can feed it the same seed phrase you have in your MetaMask, and the test network will give each of your first 10 accounts 100 test ether, which makes it easier to start work.<br />
+Using the same Mnumonic phrase will generate the same set of 10 account numbers when you restart Ganache
+``` ganache-cli -m "random set of words to generate same account numbers each restart" ```
+**Example of a Mnemonic generated by Ganache**
 
-Ganache has some great features for starting your application with different states. If your application starts with the `-m` flag, you can feed it the same seed phrase you have in your MetaMask, and the test network will give each of your first 10 accounts 100 test ether, which makes it easier to start work.
+![image](https://user-images.githubusercontent.com/1038445/150917008-2a811b12-44f8-46b7-bf7b-b6907652fe31.png)
```
mcmire, line 73: I think we can make this section more concise. We advise that people use the same seed phrase in your MetaMask, and we say if you do this it will give your first 10 accounts 100 test ether. I don't think it's necessary to explain what `-m` does, as once you give an example it will become obvious. I also don't think it's necessary to say "make sure you use the same seed phrase" every time you start Ganache, unless we feel like the reader doesn't understand that a seed phrase is used to generate the account ids (if that's the case then maybe we could link to an article that explains more about BIP39). You will _need_ to use the same seed phrase that you've used to create accounts in MetaMask, otherwise your accounts in MetaMask won't have a balance.

Putting this together, what if we say:

``` suggestion
If you start `ganache-cli` with `-m` and provide the same seed phrase you've used to create your wallet in MetaMask, the first 10 accounts in your wallet will have 100 test ether.
```

docs/guide/getting-started.md
```diff
@@ -26,11 +26,51 @@ if (typeof window.ethereum !== 'undefined') {
...
+
 ### Running a Test Network
 
-In the top right menu of MetaMask, select the network that you are currently connected to. Among several popular defaults, you'll find `Custom RPC` and `Localhost 8545`. These are both useful for connecting to a test blockchain, like [ganache](https://www.trufflesuite.com/ganache). You can quickly install and start Ganache if you have `npm` installed with `npm i -g ganache-cli && ganache-cli`.
+If you would like to run your own Network on you local machine. [Ganache](https://www.trufflesuite.com/ganache) has some great features for starting your application with different states.
+
+You can quickly install and start Ganache if you have `npm` installed with `npm i -g ganache-cli && ganache-cli`.
+Once Ganache is started connect to network `Localhost 8545` at the bottom of the network list (if you don't see `Localhost 8545` in your network list follow the instructions above [`Connecting to a test network`](#Connecting) to `show` test networks).
+You can now 
+![image](https://user-images.githubusercontent.com/1038445/150916135-c9d019ad-86f3-474d-a0ef-d53dabe799f9.png)
```
mcmire, line 66: I don't feel like this screenshot adds much value, especially when connected to the next paragraph about using `-m`. What do you think about removing it?

mcmire (changes requested):
There are some good updates here, but I think we need to make sure to use the same writing style as the rest of the document. I get that you may have tried to follow this document and wrote some things down that you found to be helpful, but some of your updates here end up being a bit confusing, so I think we need to make sure that they fit in the surrounding context. I tried to make some suggestions, let me know what you think.

madforce:
Thanks for all the feedback @mcmire. I have implemented all the recommendations. Feel free to review and update anything you think would improve the doco.

Cheers
Marty

BboyAkers:
@mcmire Mind if you take another review kind sir?? 🙂

alexandratran:
The [get started docs](https://docs.metamask.io/wallet/category/get-started) have been significantly re-designed and updated, including instructions for running a dev network.

madforce:
Awesome :)

PR #422, opened 2022-03-09
fix(example): sending transaction

sambacha:
do not send to hacked accounts even as an example

see https://github.com/MetaMask/test-dapp/pull/129/files 

cc @danfinlay


docs/guide/sending-transactions.md
```diff
@@ -56,10 +56,11 @@ sendEthButton.addEventListener('click', () => {
...
           from: accounts[0],
-          to: '0x2f318C334780961FB129D2a6c30D0763d9a5C970',
-          value: '0x29a2241af62c0000',
-          gasPrice: '0x09184e72a000',
-          gas: '0x2710',
+          to: '0x0c54FcCd2e384b4BB6f2E405Bf5Cbc15a017AaFb',
+          value: '0x0',
+          gasLimit: '0x5028',
+          gasPrice: '0x2540be400',
+          type: '0x0',
```
mcmire, line 63: While we're making this change, should we take the opportunity to send an EIP-1559 transaction here instead of a type-0 transaction?

mcmire (commented):
I don't have context on this exactly, but if we were motivated to make this change to the test dapp, then it makes sense to make the same change here. That said I left a comment.

CharlesBlaqk:
We'd be happy to look into this for you. kindly file a support ticket below 
https://forms.gle/Wmpm6NLzYm9uXYuU9

alexandratran (approved):
Updated to new docs and changed example to EIP1559 transaction.

BelfordZ:
we should document both formats, eip 1559 & pre-1559, since they are both supported.
Should we be providing a real address, as well as a real, non-zero value? Perhaps we can use placeholder values there instead?
Also, we don't need to add the field 'type' since we aren't using that in metamask as of right now.

alexandratran:
Added placeholder values, removed `type`, and added comment that legacy txns are supported. This page is just providing an example implementation of a send ETH button, not sure if adding a full legacy example here is necessary. Perhaps all legacy parameters should be documented and described in the RPC reference instead? @BelfordZ

PR #425, opened 2022-03-15
Revert "Update CircleCI config to reflect branch rename"

ghasemghafari:
[#Reverts](

- [ ] url

) MetaMask/metamask-docs#356


Gudahtt:
This is not something we want to revert

PR #476, opened 2022-05-10
Add `snap_notify` documentation

GuillaumeRx:
Fixes https://github.com/MetaMask/snaps-skunkworks/issues/418

### DIFF
--- a/docs/guide/snaps-rpc-api.md
+++ b/docs/guide/snaps-rpc-api.md
@@ -596,3 +596,70 @@ await wallet.request({
   params: ['clear'],
 });
 ```
+
+### `snap_notify`
+
+::: warning Only Callable By
+
+- Snaps
+  :::
+
+#### Parameters
+
+```typescript
+interface SnapNotifyParams {
+  /**
+   * Enum type to determine notification type.
+   */
+  type: NotificationType;
+
+  /**
+   * A message to show on the notification.
+   */
+  message: string;
+}
+```
+
+- `Array`
+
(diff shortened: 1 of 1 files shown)


docs/guide/snaps-rpc-api.md
```diff
@@ -596,3 +596,70 @@ await wallet.request({
...
+}
+```
+
+#### Returns
+
+`null` - This method doesn't return any data.
+
+#### Description
+
+Calling this method causes a notification to be added in the MetaMask UI.
```
FrederikBolding, line 650: Only `inApp` notifications are shown in the UI. Can we distinguish here?

FrederikBolding: "Calling this method shows a notification natively in the browser or in MetaMask" or something like that?

docs/guide/snaps-rpc-api.md
```diff
@@ -596,3 +596,70 @@ await wallet.request({
...
+
+```typescript
+enum NotificationType {
+  /**
+   * A default browser notification
+   */
+  native = 'native',
+
+  /**
+   * A in-app notification stored in the metamask state.
```
FrederikBolding, line 637: I think "An in-app notification showed inside the MetaMask UI" might be better, remove the line below.

FrederikBolding:
To be merged when Flask ships this commit: https://github.com/MetaMask/metamask-extension/commit/b599035a12010e31b85c3733fb573ba05b840b3c

FrederikBolding (dismissed):

rekmarks (approved):

PR #503, opened 2022-06-19
Update create-dapp.md

Monir-Nassan:
this should be enabled as the display attribute of the button is set to false


mcmire (approved):
Makes sense to me.

PR #512, opened 2022-07-01
Update create-dapp.md

Monir-Nassan:
if we are disabling the button, then I think the disabled property  should be set to true


mcmire (approved):
Makes sense to me.

PR #527, opened 2022-07-27
Update registering-your-token.md

FodiesOrg:
add logo Fodies


Gudahtt:
It looks like you directed this PR at the wrong repo. This is for the MetaMask docs, we don't want to show your logo here.

PR #531, opened 2022-08-04
Error fix in newer versions of web3

matt5346:
https://ethereum.stackexchange.com/questions/66202/undefined-results-for-web3-eth-accounts-in-truffle-console

in newer versions of web3 should be getAccounts()

### DIFF
--- a/docs/guide/signing-data.md
+++ b/docs/guide/signing-data.md
@@ -125,7 +125,7 @@ Below is an example of signing typed data with MetaMask. Reference [here](https:
 ::: tab JavaScript
 
 ```javascript
-signTypedDataV4Button.addEventListener('click', function (event) {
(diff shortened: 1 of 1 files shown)


docs/guide/signing-data.md
```diff
@@ -197,16 +197,16 @@ signTypedDataV4Button.addEventListener('click', function (event) {
     },
   });
 
-  var from = web3.eth.accounts[0];
+  var from = web3.eth.getAccounts();
```
digiwand, line 200: Hi @matt5346 👋  Thanks for the update here. One thing, `getAccounts()` returns a promise so we'll need to wait for the promise to resolve before we can get `from[0]`

digiwand (approved):
👍

PR #533, opened 2022-08-08
Update rpc-api.md

zzmp:
Include note about allowlisting the extension origin when switching chains.

### DIFF
--- a/docs/guide/rpc-api.md
+++ b/docs/guide/rpc-api.md
@@ -344,7 +344,8 @@ MetaMask stringently validates the parameters for this method, and will reject t
 if any parameter is incorrectly formatted.
 In addition, MetaMask will automatically reject the request under the following circumstances:
 
-- If the RPC endpoint doesn't respond to RPC calls from the extension's background page (if you use an origin allowlist, `chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn` must be allowed).
+- If the RPC endpoint doesn't respond to RPC calls.
+  NOTE: Calls are made from the extension's background page, _not_ the foreground page. If you use an origin allowlist, they will be blocked.
 - If the RPC endpoint returns a different chain ID when `eth_chainId` is called.
 - If the chain ID corresponds to any default MetaMask chains.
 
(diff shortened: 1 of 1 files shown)


zzmp:
I'm not sure if services (like infura) allow allowlisting the chrome-extension:// protocol (although it accepted it into the list when I tried). I'm also not sure that the request is sent from that origin, although I know it is sent from the background page (because that's where I observed this error).

It may be necessary to instead use a publicly accessible JSON-RPC URL when adding a chain, so that MetaMask's call to eth_chainId (which seems to be made to verify the URL before adding the chain) does not fail.

Gudahtt:
Thanks for the contribution! Unfortunately I'm not sure we should include this advice, because anyone implementing such an allowlist would end up blocking all Firefox users. Firefox extensions do not have any static origin (it gets generated at install-time), so there is no equivalent origin that could be used to allow just the legitimate Firefox extension.

Gudahtt:
Well, I guess we could list `moz-extension://` for Firefox, allowing _all_ traffic from Firefox extensions. Not ideal but that might work for some people. I don't know if Infura's origin allowlist would accept an entire scheme.

zzmp:
I think it's more important to note the issue than to think about whether a specific provider would allow a solution to it, because otherwise you'll just have very confused users who think that MetaMask is broken.

Is there a reason MetaMask makes this request from the background page instead of from the page trying to switch the chain? Or does it need to do so so that the popup will be able to interact with that chain without the page that originally added it?

Sawanee250825:
# p

Sawanee250825:
] Update rpc-api.md (PR #533)

Gudahtt:
The "switch chain" method doesn't interact with the network at all. The "current chain" is purely a setting internal to MetaMask. The failure you're seeing is probably the background calling `eth_chainId` to verify the chain ID with the network.

We make that request from the background because _all_ network traffic goes through the background. We route all traffic through this one process to reduce overall network traffic for users that are using multiple dapps.

Gudahtt:
I agree we should have some warning here for users about the origin allowlist. At the very least we should warn that it can cause problems. I'm just hesitant about being more specific than that without doing more testing, across Firefox and Chrome.

zzmp (commented):
docs: clarify wallet_addEthereumChain edge case

zzmp:
@Gudahtt I backed out the specificity so that this is now just a general warning with a note on the mechanism.

Gudahtt (approved):
LGTM, thanks!

PR #544, opened 2022-09-11
Update metamask-extension-provider.md

mindofmar:
The current page is pointing to old repositories and code examples so I updated it to point to the current [@metamask/providers](https://www.npmjs.com/package/@metamask/providers) package which has the most up-to-date implementations.

Feel free to add more depth, I just added what I could based on referencing other docs and an example for External Extension Provider


BboyAkers (approved):
LGTM! @Gudahtt any suggestions?

alexandratran (approved):
Updated docs and applied changes to the updated "Access a user's provider" page.

PR #546, opened 2022-09-12
upd: removed rimble ui

JoeyKhd:
It appears that the domain consensys.design is no longer operational as it was and is now put up for sale. Better to get it out of the official documentation.


Gudahtt (approved):
Good catch, thanks!

PR #553, opened 2022-09-26
Suggested improvements to Snaps getting started docs

GuiBibeau:
## Description:

This PR adds a few small improvements ton the Snaps Docs:

- makes some local host information as a link to make it clickable for users 
- makes the NVM part optional to remove time to first code!

### DIFF
--- a/docs/guide/snaps.md
+++ b/docs/guide/snaps.md
@@ -94,20 +94,12 @@ From the root of the repo, install the dependencies:
 
 ```shell
 yarn
-# or
-npm install
-# or
-pnpm install
 ```
 
 Start the development server:
 
 ```shell
 yarn start
-# or
-npm run start
-# or
-pnpm start
 ```
 
 You should now be serving both (1) the front-end and (2) the snap locally. Time to check it out in action at [`http://localhost:3000/`](http://localhost:3000/).
--- a/docs/guide/snaps.md
+++ b/docs/guide/snaps.md
@@ -82,25 +82,35 @@ Let's extend the functionality of MetaMask and build the wallet experience of th
 ### Prerequisites
 
 - Up-to-date Chromium or Firefox browser
-- [Node.js 16.0](https://nodejs.org/) (we recommend using [nvm](https://github.com/nvm-sh/nvm))
+- [Node.js](https://nodejs.org/)
 
 ### Quick start using our template
 
 Get started with Snaps using [our template](https://github.com/MetaMask/template-snap-monorepo) built with TypeScript and React. Create the repository [via GitHub](https://github.com/MetaMask/template-snap-monorepo/generate) and [clone it](https://docs.github.com/en/repositories/creating-and-managing-repositories/cloning-a-repository) down to your local machine using e.g. the command line.
 
-To ensure the correct Node version, `cd` into your cloned repository and run:
+> NB: Snaps should work with the latest LTS version of Node.js, but we recommend using the version specified in the template's `.nvmrc` file. If you use [nvm](https://github.com/nvm-sh/nvm) you can switch easily with calling `nvm use` at the root of the project.
+
+From the root of the repo, install the dependencies:
 
 ```shell
-nvm use
+yarn
+# or
+npm install
+# or
+pnpm install
 ```
 
-Setup the development environment:
+Start the development server:
 
 ```shell
-yarn install && yarn start
+yarn start
+# or
+npm run start
+# or
+pnpm start
 ```
 
-You should now be serving both (1) the front-end and (2) the snap locally. Time to check it out in action at `http://localhost:3000/`.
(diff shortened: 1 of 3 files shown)


docs/guide/snaps.md
```diff
@@ -82,25 +82,35 @@ Let's extend the functionality of MetaMask and build the wallet experience of th
...
 - Up-to-date Chromium or Firefox browser
-- [Node.js 16.0](https://nodejs.org/) (we recommend using [nvm](https://github.com/nvm-sh/nvm))
+- [Node.js](https://nodejs.org/)
 
 ### Quick start using our template
 
 Get started with Snaps using [our template](https://github.com/MetaMask/template-snap-monorepo) built with TypeScript and React. Create the repository [via GitHub](https://github.com/MetaMask/template-snap-monorepo/generate) and [clone it](https://docs.github.com/en/repositories/creating-and-managing-repositories/cloning-a-repository) down to your local machine using e.g. the command line.
 
-To ensure the correct Node version, `cd` into your cloned repository and run:
+> NB: Snaps should work with the latest LTS version of Node.js, but we recommend using the version specified in the template's `.nvmrc` file. If you use [nvm](https://github.com/nvm-sh/nvm) you can switch easily with calling `nvm use` at the root of the project.
```
mcmire, line 91: I rather like this a lot, streamlines the instructions without removing this information 👏🏻

eriknson: Indeed, I like this direction.

docs/guide/snaps.md
```diff
@@ -82,25 +82,35 @@ Let's extend the functionality of MetaMask and build the wallet experience of th
 ### Prerequisites
 
 - Up-to-date Chromium or Firefox browser
-- [Node.js 16.0](https://nodejs.org/) (we recommend using [nvm](https://github.com/nvm-sh/nvm))
+- [Node.js](https://nodejs.org/)
```
FrederikBolding, line 85: Is there a reason not to specify a minimum Node version?

docs/guide/snaps.md
```diff
@@ -82,25 +82,35 @@ Let's extend the functionality of MetaMask and build the wallet experience of th
...
-To ensure the correct Node version, `cd` into your cloned repository and run:
+> NB: Snaps should work with the latest LTS version of Node.js, but we recommend using the version specified in the template's `.nvmrc` file. If you use [nvm](https://github.com/nvm-sh/nvm) you can switch easily with calling `nvm use` at the root of the project.
+
+From the root of the repo, install the dependencies:
 
 ```shell
-nvm use
+yarn
+# or
+npm install
```
FrederikBolding, line 98: Have you verified that this works and is smooth? 😅 - I wonder if the experience is fine using NPM instead of Yarn given that the repo is set up to use with Yarn.

docs/guide/snaps.md
FrederikBolding: FYI Snaps will not work with <16

GuiBibeau: From reading the docs the first time, it felt like a lot of information to take in in order to get started. 

 I wanted to strike a balance with making information available but not overwhelming people with a big cognitive load. I see two cases that can happen here if a user is not on 16+:
 
 - Experienced users will see the `.nvmrc` file and adjust themselves quickly
 - Less experienced users will revert back here and stumble on the `<blockquote>` explaining the exact information I am removing here.

docs/guide/snaps.md
GuiBibeau: I'll test out the full flow and revert! I expect it to be equivalent as the stated goal of Yarn was always full interchangeability. But better be safe than sorry 😅 thanks for raising this!

mcmire: > I expect it to be equivalent as the stated goal of Yarn was always full interchangeability

That may have been _a_ stated goal back in the Yarn 1 days, but I don't think interoperability is a goal now. Both npm and Yarn have some similar features, but there is no guarantee that they will work well with each other (e.g. workspaces). Also, as @FrederikBolding said, the repo definitely depends on Yarn — the scripts within the repo won't run unless you have Yarn installed. So I would expect this not to work. May be better to revert as you say, and also list Yarn as a prerequisite, in case the user doesn't have that installed.

GuiBibeau: I opened up the template and noticed indeed that we leverage yarn workspaces. I updated this change out to use only `yarn` so that my changes don't break the tutorial

As a side note to explain my initial intentions:

I have some concerns about adopting an opinionated approach towards a package manager vs another one when we are building developer tooling and tutorials. This looks fine for a beta but users will have their own toolset that they wish to bring with them and if there is a way to be inclusive with that toolset I would strive towards supporting all package managers including yarn, npm and pnpm.

I understand this is a bigger development effort but it should be a standard we impose on our self.

Mrtenz (approved):

httpJunkie (approved):
Looks good!

PR #555, opened 2022-10-03
Add docs about `snap_getBip32PublicKey`

Mrtenz:
This adds docs about the new `snap_getBip32PublicKey` method.

Fixes https://github.com/MetaMask/snaps-monorepo/issues/802

### DIFF
--- a/docs/guide/snaps-rpc-api.md
+++ b/docs/guide/snaps-rpc-api.md
@@ -554,6 +554,75 @@ const addressKey1 = await deriveDogecoinAddress(1);
 // Now, you can ask the user to e.g. sign transactions!
 ```
 
+### `snap_getBip32PublicKey`
+
+::: warning Only Callable By
+
+- Snaps
+  :::
+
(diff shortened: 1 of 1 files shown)


docs/guide/snaps-rpc-api.md
```diff
@@ -554,6 +554,75 @@ const addressKey1 = await deriveDogecoinAddress(1);
...
+::: tab Code
+
+```javascript
+// By way of example, we will use Dogecoin, which has a derivation path starting
+// with `m/44'/3'`.
+const dogecoinPublicKey = await wallet.request({
+  method: 'snap_getBip32PublicKey',
+  params: {
+    // The path and curve must be specified in the initial permissions.
+    path: ['m', "44'", "3'", "0'", '0', '0'],
```
FrederikBolding, line 612: In this current implementation this wouldn't work with the above permissions.

FrederikBolding (approved):

PR #576, opened 2022-11-08
Fix docs: Change port from `3000` to `8000`

strykerin:
When running `yarn start` the port is `8000` instead of `3000`


FrederikBolding (approved):

FrederikBolding:
Thanks for catching this!

PR #619, opened 2022-12-21
Make Flask a prerequisite before running Snaps (#599)

### LINKED ISSUE #599: Reorder the instructions in the Snaps introduction to introduce Flask as a pre-requisite
There’s a break between when we tell users to navigate to localhost:8000 and when we tell them to press the “Connect” button. This weirdness is partly because we introduce MetaMask Flask after telling them to use the template. I think the solution is to tell them to install MetaMask Flask in the pre-requisites and then the steps to use the template will flow easily. 

As an aside, we will not have to tell people to refresh to detect Flask either.

ziad-saab:
Tell users to install Flask before running the example. This way a browser refresh won't be necessary, and less confusion will result. Fixes #599

### DIFF
--- a/docs/guide/snaps.md
+++ b/docs/guide/snaps.md
@@ -83,6 +83,13 @@ Let's extend the functionality of MetaMask and build the wallet experience of th
 
 - Up-to-date Chromium or Firefox browser
 - [Node.js](https://nodejs.org/)
+- MetaMask Flask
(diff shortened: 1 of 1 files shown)


Montoya (approved):

bowensanders (approved):
LGTM

PR #623, opened 2022-12-22
Improve snaps documentation (#275)

ziad-saab:
This PR makes multiple small improvements to the Snaps documentation. It fixes https://github.com/MetaMask/MetaMask-planning/issues/275 .


GuillaumeRx (commented):
LGTM to me but we should hold of on releasing this since it talks about things that aren't live in Flask yet.

docs/guide/snaps-development-guide.md
```diff
@@ -250,16 +261,20 @@ Via the MetaMask settings page, the user can see their installed snaps. For each
...
 Providing more ways for snaps to modify the MetaMask UI is an important goal of the snaps system, and over time more and more snaps will be able to contain their user interfaces entirely within MetaMask itself.
 
 #### Detecting the User's MetaMask Version
 
-When developing a website that depends on Snaps, it's important to know whether MetaMask Flask is installed.
+When developing a website/Dapp that depends on Snaps, it's important to know whether MetaMask Flask is installed.
 For this purpose, we recommend using the [`@metamask/detect-provider`](https://npmjs.com/package/@metamask/detect-provider)
-package [`web3_clientVersion`](https://metamask.github.io/api-playground/api-documentation/#web3_clientVersion)
+package's [`web3_clientVersion`](https://metamask.github.io/api-playground/api-documentation/#web3_clientVersion)
 RPC method as demonstrated in the following snippet:
```
mcmire, line 278: This makes it sound like the `@metamask/detect-provider` package provides the `web3_clientVersion` RPC method, but that's not true — the functionality is actually coming from the provider via MetaMask. Maybe this should say:

``` suggestion
We recommend calling the [`web3_clientVersion`](https://metamask.github.io/api-playground/api-documentation/#web3_clientVersion) RPC method to obtain this information. Here we are using the [`@metamask/detect-provider`](https://npmjs.com/package/@metamask/detect-provider) package to get the provider object from MetaMask first:
```

Alternatively, at the top of this document, in "The Snap Source Code", there is an example of using `ethereum.request` to make a request. Why isn't `detect-provider` being used there? If we know we are talking to MetaMask, why can't we use `ethereum.request` directly here?

mcmire (commented):
Made a few comments/suggestions.

docs/guide/snaps-development-guide.md
```diff
@@ -67,6 +75,7 @@ Consider this trivial snap, which we'll call `hello-snap`:
 ```javascript
 module.exports.onRpcRequest = async ({ origin, request }) => {
   switch (request.method) {
+    // Expose a "hello" RPC method to dApps
```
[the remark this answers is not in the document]
ziad-saab: Looks like we spell it "Dapp" in other places. I'll change.

docs/guide/snaps-development-guide.md
```diff
@@ -250,16 +261,20 @@ Via the MetaMask settings page, the user can see their installed snaps. For each
...
+- By returning transaction insights from the [`onTransaction`](./snaps-exports.md#ontransaction) export.
+
+This means that most snaps will have to rely on web pages (i.e., dapps) and their own RPC methods to present their data to the user.
 
 Providing more ways for snaps to modify the MetaMask UI is an important goal of the snaps system, and over time more and more snaps will be able to contain their user interfaces entirely within MetaMask itself.
 
 #### Detecting the User's MetaMask Version
 
-When developing a website that depends on Snaps, it's important to know whether MetaMask Flask is installed.
+When developing a website/Dapp that depends on Snaps, it's important to know whether MetaMask Flask is installed.
```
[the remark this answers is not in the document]
ziad-saab: Dapp seems to be the winner...

docs/guide/snaps-development-guide.md
```diff
@@ -14,7 +14,15 @@ Read on to learn more!
...
 
 Before continuing, you should know that [`@metamask/snaps-cli`](https://www.npmjs.com/package/@metamask/snaps-cli) exists, and will be one of your most important tools as you get started with snap development.
-The CLI can be installed globally using `npm` or `yarn`, and provides commands for initiating a snap project and building, executing, and serving your snap for local development.
+The CLI can be installed globally using `npm` or `yarn`:
+
+```sh
+yarn global add @metamask/snaps-cli
+# or...
+npm i -g @metamask/snaps-cli
+```
```
Mrtenz, line 23: Local installation makes it easier for other developers working on the same project to use the CLI (through scripts), so I think we should prefer that over global installation.

Montoya: If we want to recommend that people use the CLI to create new projects, how would they do that?

ziad-saab: How about having a separate package `create-metamask-app` that can [work with `yarn create`](https://classic.yarnpkg.com/blog/2017/05/12/introducing-yarn/), this would be the equivalent of the current `mm-snap init`. And then `mm-snap` is a local install that just deals with the management of the current snap?

mcmire: I like that idea @ziad-saab.

In the meantime, given that this seems to be the current process, maybe it's okay to merge this as-is?

Mrtenz (changes requested):

Mrtenz (approved):

FrederikBolding (approved):

PR #658, opened 2023-02-09
Update signing-data.md based on eth_sign behavior changes

### LINKED ISSUE #657: Update guidelines around eth_sign based on planned change in MetaMask Extension behavior
Make copy adjustments based on the change to behavior of the eth_sign API method that's been implemented as part of https://github.com/MetaMask/metamask-extension/issues/17491

vandan:
Minor changes to resolve #657

I made the changes in the markdown, built, and verified the rendering on localhost.
While there is one other mention of eth_sign I found under ethereum-json-rpc-methods, I didn't think that removing it or adding a deprecation notice made sense there because we still have a valid implementation of the Ethereum eth_sign method (I could be wrong).

This is my first MetaMask PR so please give it some extra scrutiny.

### DIFF
--- a/docs/guide/signing-data.md
+++ b/docs/guide/signing-data.md
@@ -4,13 +4,15 @@ MetaMask lets you request cryptographic signatures from users in a number of way
 
 - [eth_signTypedData_v4](#signtypeddata-v4) for the most readable signatures that are also efficient to process on chain.
 - [personal_sign](#personal-sign) for the easiest way to get a human readable signature of data that does not need to be efficiently processed on-chain.
-- `eth_sign` (Not recommended) If you need to perform low-level signatures and don't mind having the user presented with high security friction, because the proposal is not readable.
+- `eth_sign` (deprecated) By default MetaMask treats this method as disabled because the proposal is not readable.
 
 | Signing Methods        | Human readable | Efficient to process on-chain | Easy to use |
 | ---------------------- | -------------- | ----------------------------- | ----------- |
 | `eth_signTypedData_v4` | ✅             | ✅                            | ❌          |
 | `personal_sign`        | ✅             | ❌                            | ✅          |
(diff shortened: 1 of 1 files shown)


BelfordZ:
Nice work @vandan 

Could you reference a PR or issue in extension which adds the setting to enable/disable eth_sign? What is the plan for rolling this out?

Is there an api-specs PR / issue as well?

vandan:
Adding the reference to the issue in the extension repo for the actual change in behavior when eth_sign calls are made: https://github.com/MetaMask/metamask-extension/issues/17491

@BelfordZ I believe this change is being rolled out imminently (with Extension v10.25.0). Because eth_sign isn't currently being included in the API spec, I didn't think changing the API spec would be necessary at this stage.

docs/guide/signing-data.md
```diff
@@ -287,7 +289,7 @@ If you’d like to read our JavaScript implementations of these methods, they ar
 
 Note that MetaMask supports signing transactions with Trezor and Ledger hardware wallets. These hardware wallets currently only support signing data using the `personal_sign` method. If you have trouble logging in to a website or dapp when using a Ledger or Trezor, the site may be requesting you sign data via an unsupported method, in which case we recommend using your standard MetaMask account.
 
-## A Brief History
+## <a name="history"></a> A Brief History
```
FrederikBolding, line 292:
```suggestion
## A Brief History
```
I don't think this is necessary, given that we generate anchor links. You should probably update the link above to `#a-brief-history`

vandan: Ah, I wasn't quite sure how to do that. I'll try out the suggested anchor link instead.

vandan: Worked like a charm!

alexandratran (approved):

FrederikBolding (approved):

shanejonas (approved):

PR #660, opened 2023-02-11
Fixed typo in "Sending Transaction" error handler

cuamckuu:
Hi, I found a typo while playing with example from ["Sending Transactions" page](https://docs.metamask.io/guide/sending-transactions.html#example)

Current version of error handler does nothing:
```javascript
.catch((error) => console.error);
```

But it probably should print an error, and be like:
```javascript
.catch((error) => console.error(error));
```


BboyAkers (approved):
Thanks for catching this!

PR #722, opened 2023-04-20
Add Usabilla feedback script

vandan:
(Edited by @alexandratran)

Added feedback survey to the SDK doc pages. It's configured to target only a specific subset of returning users, so likely won't appear on the PR preview.

https://github.com/MetaMask/metamask-docs/assets/12214231/0e87b387-5dde-45a4-aab9-490f60a440e5


vandan:
The custom-script.js doesn't seem to get deployed properly for preview builds.

Even once that's resolve, we're just previewing for now. NO MERGE should be done until we confirm whether the addition of a cookie consent widget is a prerequisite.

docusaurus.config.js
```diff
@@ -29,6 +29,7 @@ const config = {
 
   scripts: [
     { src: "https://plausible.io/js/script.js", defer: true, "data-domain": "docs.metamask.io" },
+    { src: "/js/custom-script.js", defer: true, async: true },
```
BelfordZ, line 32:
```suggestion
    { src: `${process.env.DEST || ""}/js/custom-script.js`, defer: true, async: true },
```

vandan:
[Reference PR](https://github.com/INFURA/docs/pull/63) for Infura's implementation of similar functionality.

mjsmike62 (approved):
Based on my observation, the code seems to be formatted correctly. Also, I thought the feedback survey was done well. The only things that stuck out to me were the following (but I don't see these as anything major) ...

- Use of exclamation (!) mark in "Help improve MetaMask SDK!" Maybe the exclamation mark is a little too much? You could even say, "Would you like to help improve MetaMask SDK?"
- Statement: "Which platform are you building your dapp on?" Maybe, "On which platform are you building your dapp?"
- Statement: "What are your top 3 concerns..." Maybe, "What are your top three concerns..."
- Final window... "Thanks!  Thank you for..." Maybe, "Thanks! Your feedback is much appreciated." No need to thank the user twice.

PR #730, opened 2023-04-25
Load content

BelfordZ:


README.md
```diff
@@ -169,3 +169,16 @@ filesystem.
...
+
+Reference documentation is generated from [the docusaurus-openrpc plugin](https://github.com/MetaMask/docusaurus-openrpc) using [api-specs](https://github.com/MetaMask/api-specs/) as the OpenRPC document source. In order to update these docs, follow the following steps:
+
+1. Clone [api-specs](https://github.com/MetaMask/api-specs/)
+  1. `cd api-specs`
+  2. `yarn`
+  3. `yarn build:watch`
+2. Inside metamask-docs/docusaurus.config.js:
+  1. update the reference to the built openrpc doc under the key 'openrpcDocument'. It should point to whereever you have the api-specs repo cloned. ex: `openrpcDocument: "../api-specs/dist/build/openrpc.json"`
+  2. `yarn start`
```
alexandratran, line 184:
```suggestion
### Update RPC API documentation

Wallet RPC API reference documentation is generated from the [`docusaurus-openrpc`](https://github.com/MetaMask/docusaurus-openrpc) plugin using [`api-specs`](https://github.com/MetaMask/api-specs/) as the OpenRPC document source. To update this documentation:

1. Clone [`api-specs`](https://github.com/MetaMask/api-specs/), and run `yarn` and `yarn build:watch` in your local repository.
2. In `metamask-docs/docusaurus.config.js`:
  1. Update the reference to the built OpenRPC document under the key 'openrpcDocument'. It should point to wherever you have the `api-specs` repository cloned. For example, `openrpcDocument: "../api-specs/dist/build/openrpc.json"`.
  2. Run `yarn start`.
```

alexandratran (approved):

ziad-saab:
I see two odd things here:

1. The `eth_accounts` method appears twice >> https://docs.metamask.io/load-content/api-playground#eth_accounts They both use the same anchor but if you cmd+f on the page you'll see the method twice.
2. The `eth_accounts` method's return value is output as `AddressList`, but then "Array of nothing":
   image
   I think it's possibly because in the schema we're using:
   image
   The `AddressList` is defined using a `$ref`, and that's not being handled properly by the plugin we're using.

BelfordZ:
closed in favor of https://github.com/MetaMask/metamask-docs/pull/763

PR #745, opened 2023-04-30
Update instructions for creating a snap

### LINKED ISSUE #775: `yarn global add` is not supported in yarn v2 or v3
[In the docs for the snaps-cli](https://docs.metamask.io/snaps/get-started/install-snaps/#install-the-snaps-cli) we tell users to use npm or yarn to install the snaps cli globally, but this has been deprecated in yarn: https://yarnpkg.com/getting-started/migration#use-yarn-dlx-instead-of-yarn-global

We should just tell users to use npm and drop the mention of yarn, or include instructions for using `yarn dlx`.

ziad-saab:
Snaps should be created using the CLI rather than cloning the repo.

This should only be merged when the `@metamask/create-snap` package is published on NPM through this PR: https://github.com/MetaMask/snaps-monorepo/pull/1268

fixes #775

### DIFF
--- a/snaps/get-started/install-snaps.md
+++ b/snaps/get-started/install-snaps.md
@@ -4,7 +4,7 @@ description: Install Snaps dependencies.
 
 # Install Snaps
 
-To use Snaps, you must install [MetaMask Flask](#install-metamask-flask) and the [Snaps CLI](#install-the-snaps-cli).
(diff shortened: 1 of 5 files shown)


alexandratran (approved):
Made some minor edits, LGTM.

PR #763, opened 2023-05-15
Add OpenRPC Plugin for API Reference docs/playground

shanejonas:


BelfordZ:
todo:

 - [ ] address some of the socket security vulns
 - [ ] pass docusaurus' MdxRenderer into open-rpc-docs-react to keep markdown rendering unified
 - [ ] allow openrpc extension for overriding the default transport/example template
 - [ ] better styling for json-schema-form usage (aligment of buttons)
 - [ ] json-schema improvements to api-specs
   - [ ] cant edit json schema with type object and no properties (wallet_requestPermissions)
   - [ ] methods that are inherited from execution-apis should have a tag to distinguish them
   - [ ] wallet_watchAsset
     - [ ] has a link in the description that is not clickable.
     - [ ] constrain the number for decimals to "must be between 0 and 36"
     - [ ] image should have the format: 'uri'
     - [ ] address should ref the address schema (which should include a regex and length requirements, etc)
   - [ ] method content descriptors should have descriptions (and summary where applicable).
   - [ ] all methods should have at least one examplePairing
   - [ ] filter out execution api methods we dont support in MM dapp api (eth_createAccessList for example)
 - [ ] not enough spacing between method name and the method description

PR #768, opened 2023-05-23
Fix typo in gas-estimation.md

yuhatown:
Alert -> alert

An error happened: The "type" property must be one of: alert, confirmation, prompt.


FrederikBolding (approved):
Good catch - thanks!

PR #771, opened 2023-05-25
Add information about same-origin policy and CORS in snaps

### LINKED ISSUE #579: Document `fetch()` CORS issue in Snaps
CORS issues when using `fetch()` in Snaps are non-obvious. We should document that they happen and how to debug them.

**AC**

Documentation has been updated.

ziad-saab:
fixes #579

### DIFF
--- a/snaps/reference/permissions.md
+++ b/snaps/reference/permissions.md
@@ -98,6 +98,10 @@ Specify this permission in the manifest file as follows:
 },
 ```
 
+### Same-origin policy and CORS
+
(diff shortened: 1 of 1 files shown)


snaps/reference/permissions.md
```diff
@@ -98,6 +98,10 @@ Specify this permission in the manifest file as follows:
 },
 ```
 
+### Same-origin policy and CORS
+
+`fetch()` requests done inside a snap are bound by browsers' [same-origin policy](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy#cross-origin_network_access). Since snaps code is executed in an iframe with the `sandbox` property, the browser will send an `Origin` header with the value `null` with outgoing requests. For the snap to be able to read the response, an [`Access-Control-Allow-Origin`](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) CORS header with the value `*` or `null` must be sent in the response.
```
alexandratran, line 103:
```suggestion
`fetch()` requests in a snap are bound by browsers' [same-origin policy](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy#cross-origin_network_access). Since snap code is executed in an iframe with the `sandbox` property, the browser sends an `Origin` header with the value `null` with outgoing requests. For the snap to be able to read the response, the server must send an [`Access-Control-Allow-Origin`](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) CORS header with the value `*` or `null` in the response.
```

alexandratran (approved):
Minor suggestions.

PR #801, opened 2023-06-08
Updated to move deprecated methods to later section

### LINKED ISSUE #764: Suggestion: Divide the signing methods page into 2 sections: Recommended and Deprecated (or Current and Deprecated)
The information presented on this page is chronological and accurate but it is not the most useful way to present the information to a developer who just wants to find the best ways to use signing right now: 
https://docs.metamask.io/wallet/concepts/signing-methods

It makes sense to present accurate information but I have an idea on how this could be rewritten to prioritize giving the most useful information first. Make two sections, the first section being "recommended" or "current" (not sure what the best term would be) and the second section "deprecated." 

So the first section would have two subheads, 1. personal_sign and 2. eth_signTypedData_v4, and the second would have everything else with a big disclaimer that all these methods are not recommended to be used.

MadelineMurray:
fixes #764 

As suggested in issue, moved current recommended methods to top of article and other methods to later section titled deprecated.

### DIFF
--- a/wallet/concepts/signing-methods.md
+++ b/wallet/concepts/signing-methods.md
@@ -1,72 +1,75 @@
 ---
-description: A brief history of the signing RPC methods.
+description: Available RPC methods for signing transactions in MetaMask.
 sidebar_position: 2
 ---
 
 # Signing methods
 
-This page describes a brief history of the signing RPC methods in MetaMask.
+This page describes the signing RPC methods in MetaMask.
 Learn how to [use the recommended signing methods](../how-to/sign-data.md).
 
-### eth_sign
+### eth_signTypedData_v4
 
-`eth_sign` is MetaMask's original signing method.
-It allows signing an arbitrary hash, which means it can be used to sign transactions, or any other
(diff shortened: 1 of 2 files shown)


wallet/concepts/signing-methods.md
```diff
@@ -1,72 +1,75 @@
...
+MetaMask disables `eth_sign` by default and does not recommend using `eth_sign` in production.
+Some applications (usually internal administrator panels) use `eth_sign` for ease of
+use, or due to an inability to change the associated dapp.
+If a wallet user must interact with a dapp that uses `eth_sign` and accepts the risks,
+the wallet user can re-enable `eth_sign` through advanced settings.
+:::
 
-If on-chain verifiability cost is a high priority for you, we recommend
-[using this method](../how-to/sign-data.md#use-eth_signtypeddatav4).
+### eth_signTypedData_v1 and eth_signTypedData_v3
```
Montoya, line 56: I think this was supposed to be #### not ###?

alexandratran (approved):
Added minor edits. LGTM.

PR #810, opened 2023-06-13
Updated to make task focused

### LINKED ISSUE #809: Update the Introduction to focus on what integrating MM enables
The Introduction is currently focused on the technology (eg, What is the Ethereum Provider API) rather than what it enables (access to the blockchain from your dapp).

MadelineMurray:
Fixes #809 

Updated introduction to be task focused (ie, what it enables) rather than feature focused (ie, describe how it's implemented)

### DIFF
--- a/wallet/index.md
+++ b/wallet/index.md
@@ -2,13 +2,29 @@
 title: Introduction
 ---
 
-# Integrate with the MetaMask wallet
+# Integrate your dapp with MetaMask
 
-Integrate your dapp with the MetaMask wallet using the
-[MetaMask Ethereum provider API](reference/provider-api.md), which enables your dapp to interact
-with its users' Ethereum accounts.
-We recommend using [MetaMask SDK](how-to/use-sdk/index.md) to easily enable your users to connect to
-their MetaMask wallet client from any platform.
+Integrate your dapp with the MetaMask wallet to enable users to connect their MetaMask wallet from: 
+
+- Desktop and mobile browsers. 
+- Desktop and mobile applications.
+- Games built on platforms such as Unity. 
+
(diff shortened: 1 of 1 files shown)


wallet/index.md
```diff
@@ -2,13 +2,29 @@
...
+- Desktop and mobile browsers. 
+- Desktop and mobile applications.
+- Games built on platforms such as Unity. 
+
+Integrating your dapp with the MetaMask wallet takes only two lines of code and enables: 
+
+- Connecting your application to millions of MetaMask users.
+- Access to all chains supported by MetaMask. 
+
+The methods for integrating the MetaMask wallet into your dapp are:
```
alexandratran, line 18: This seems to present the SDK and provider API as mutually exclusive options, but, as I understand it, the SDK directly uses the provider API; developers can call any provider API method as usual when using the SDK. Can we reflect that in this paragraph?

MadelineMurray: @alexandratran - this is a good point.  What's your sense on introducing the concept of the MM provider API in the intro?  This is one of the things that threw me to start with when trying to understand how the MM integration works - it's diving into how (ie, using the MM Provider) rather than what (ie, connecting to MM).  But if I try and write it without mentioning the MM provider, it doesn't quite work.

alexandratran: I lean towards briefly introducing the concept of the provider API in the intro, since it provides crucial context for the SDK and can technically be used without the SDK (which is how most developers currently use it). Then maybe we can link to a separate concepts page that explains how the SDK uses the provider in more detail.

alexandratran:
Superseded by #840

PR #829, opened 2023-06-20
Update import

fibo:
I am trying **@metamask/sdk** `v^0.3.2` and it looks like the `MetaMaskSDK` is not exported as default.


deeeed:
We are slowly moving away from default exports. 
Named exports are safer (and easier to import directly from IDE) and mixing both can lead to errors.

You can import using `import {MetaMaskSDK} from '@metamask/sdk' ;`

alexandratran (approved):

PR #861, opened 2023-07-19
Add Snaps Simulator to the snaps resources list

FrederikBolding:
Add Snaps Simulator to the snaps resources list.

Could potentially also be added in other places?


Montoya (approved):
LGTM

PR #878, opened 2023-08-02
Update snaps examples linked from docs

Montoya:
Had to update some outdated links in the development guide. Sticking to only the snaps that have been developed internally and are still relevant to our current API.

### DIFF
--- a/snaps/how-to/develop-a-snap.md
+++ b/snaps/how-to/develop-a-snap.md
@@ -105,7 +105,7 @@ can always build, publish, and use snaps without MetaMask's permission.
 
 ## Resources and tools
 
-You can review the growing number of [example snaps](https://github.com/MetaMask/snaps-monorepo/tree/main/packages/examples) maintained by MetaMask, as well as the following fully functional and open source snaps: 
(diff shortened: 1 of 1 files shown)


alexandratran (approved):

snaps/how-to/develop-a-snap.md
```diff
@@ -105,13 +105,11 @@ can always build, publish, and use snaps without MetaMask's permission.
 
 ## Resources and tools
 
-You can review the growing number of
-[example snaps](https://github.com/MetaMask/snaps-monorepo/tree/main/packages/examples) maintained
```
Mrtenz, line 109: Why are we removing this?

Montoya: Are we still maintaining it? If so, we can keep it. Sorry!

Mrtenz: I redid all the examples a couple of months ago and added a lot of documentation etc., and intend to keep this up to date. We also use them for e2e testing, so all features we want to test will have an example snap, so this seems like a good place to point people to.

PR #883, opened 2023-08-09
Update permissions.md

fribas84:
missing permissions for snap_notify and snap_dialog

### DIFF
--- a/snaps/reference/permissions.md
+++ b/snaps/reference/permissions.md
@@ -158,3 +158,22 @@ Specify this permission in the manifest file as follows:
   "endowment:webassembly": {}
 },
 ```
+
+## snap_notify 
+To display a notification in MetaMask or natively in the browser, a snap must request `snap_notify` permission.
+Specify this permission in the manifest file as follows:
+
+```json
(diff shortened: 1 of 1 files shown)


alexandratran (commented):
The restricted methods that require permissions are already specified in the [JSON-RPC reference](https://docs.metamask.io/snaps/reference/rpc-api/#restricted-methods), so they don't need to be added to this page. Do you think we need to make that more clear in the docs?

fribas84:
Hi, @alexandratran Thanks for following this issue. Some enhancements can be improved that's the reason for this PR. I updated this because when I was building I faced the problem "permissions required or missing permissions (don't recall right now the correct wording)", I checked in the tab "Snap Permissions" and there was no info about the requirements that I needed, that is the reason for my change.

alexandratran:
@fribas84 Thanks for clarifying! I've raised an issue (https://github.com/MetaMask/metamask-docs/issues/884) to take a look at this further to figure out the best way to clarify snaps permissions in the docs. Will close this PR for now.

PR #886, opened 2023-08-16
Improve snaps permissions doc

alexandratran:
Add section on restricted RPC methods in Snaps permissions reference, for clarity. Fixes #883

### DIFF
--- a/snaps/reference/permissions.md
+++ b/snaps/reference/permissions.md
@@ -8,7 +8,23 @@ sidebar_position: 4
 You can [request the following permissions](../how-to/request-permissions.md) in your snap manifest
 file.
 
-## endowment:cronjob
+## RPC API permissions
+
+You must request permission to use any
+[restricted JSON-RPC API methods](rpc-api.md#restricted-methods).
+
+For example, to request to use [`snap_dialog`](rpc-api.md#snap_dialog), add the following to the
+manifest file:
+
+```json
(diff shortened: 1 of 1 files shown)


snaps/reference/permissions.md
```diff
@@ -98,11 +114,16 @@ Specify this permission in the manifest file as follows:
 },
 ```
 
-### Same-origin policy and CORS
+#### Same-origin policy and CORS
 
-`fetch()` requests in a snap are bound by browsers' [same-origin policy](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy#cross-origin_network_access). Since snap code is executed in an iframe with the `sandbox` property, the browser sends an `Origin` header with the value `null` with outgoing requests. For the snap to be able to read the response, the server must send an [`Access-Control-Allow-Origin`](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) CORS header with the value `*` or `null` in the response.
+`fetch()` requests in a snap are bound by browsers' [same-origin policy](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy#cross-origin_network_access).
```
rajkaramchedu, line 119: This might read better if we say "browser's" because a user session usually occurs on a single browser?

rajkaramchedu (approved):
I left a comment, approving.

PR #889, opened 2023-08-21
Document the eth_accounts permission for snaps

### LINKED ISSUE #888: Document the eth_accounts permission for snaps
Document the first dynamic permission for snaps: `eth_accounts`

ziad-saab:
Fixes #888

### DIFF
--- a/snaps/reference/permissions.md
+++ b/snaps/reference/permissions.md
@@ -184,7 +184,7 @@ Specify this permission in the manifest file as follows:
 
 Dynamic permissions are not requested in the manifest file. Instead, they are acquired during the lifecycle of the snap.
 
-## eth_accounts
+### eth_accounts
 
 This dynamic permission is acquired when a snap calls the `eth_requestAccounts` method of the Ethereum provider. Calling `eth_requestAccounts` requires the [`ethereum-provider` endowment](#endowmentethereum-provider). The presence of the permission can be checked by calling [`wallet_getPermissions`](/wallet/reference/rpc-api/#wallet_getpermissions). If the permission is present, the return value of `wallet_getPermissions` will contain a permission with a `parentCapability` of `eth_accounts`. It will come with a caveat of `restrictReturnedAccounts`, an array of all the accounts that the user allowed for this snap:
 
--- a/snaps/reference/permissions.md
+++ b/snaps/reference/permissions.md
@@ -179,3 +179,30 @@ Specify this permission in the manifest file as follows:
   "endowment:webassembly": {}
 },
 ```
+
+## Dynamic permissions
+
+Dynamic permissions are not requested in the manifest file. Instead, they are acquired during the lifecycle of the snap.
+
+## eth_accounts
+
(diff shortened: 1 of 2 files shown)


snaps/reference/permissions.md
```diff
@@ -179,3 +179,30 @@ Specify this permission in the manifest file as follows:
...
 },
 ```
+
+## Dynamic permissions
+
+Dynamic permissions are not requested in the manifest file. Instead, they are acquired during the lifecycle of the snap.
+
+## eth_accounts
+
+This dynamic permission is acquired when a snap calls the `eth_requestAccounts` method of the Ethereum provider. Calling `eth_requestAccounts` requires the [`ethereum-provider` endowment](#endowmentethereum-provider). The presence of the permission can be checked by calling [`wallet_getPermissions`](/wallet/reference/rpc-api/#wallet_getpermissions). If the permission is present, the return value of `wallet_getPermissions` will contain a permission with a `parentCapability` of `eth_accounts`. It will come with a caveat of `restrictReturnedAccounts`, an array of all the accounts that the user allowed for this snap:
```
FrederikBolding, line 189: Worth mentioning that you should be able to request `eth_accounts` and/or `eth_requestAccounts`, which will also return the addresses without the extra permission caveats information?

snaps/reference/permissions.md
```diff
@@ -179,3 +179,30 @@ Specify this permission in the manifest file as follows:
...
+      "value": [
+        "0xc403b37bf1e700cb214ea1be9de066824b420de6" // example connected account #1
+      ]
+    }
+  ],
+  "date": 1692616452846
+}
+```
+
+This permission can be revoked by the user by going to the "Connected sites" option in the MetaMask UI, and removing the connection from an account to the snap.
```
FrederikBolding, line 208: This needs to be revoked in the specific snap's settings under snap permissions.

FrederikBolding (changes requested):

snaps/reference/permissions.md
ziad-saab: @FrederikBolding do you mean using `wallet_requestPermissions`? I didn't cover that in the PR, can't find any info on how to use it. My understanding is that calling `eth_accounts` won't popup the MetaMask window, so it won't generate any new permission. Can you elaborate on this please?

FrederikBolding: Sorry, I mean after calling `eth_requestAccounts` to get the `eth_accounts` permission, you can simply call it again to get the already granted accounts. You don't need to use `wallet_getPermissions`. IIRC requesting `eth_accounts` should do the same. But fine by me to just reference `eth_requestAccounts`, it feels like a better pattern than `wallet_getPermissions`

snaps/reference/permissions.md
```diff
@@ -179,3 +179,30 @@ Specify this permission in the manifest file as follows:
   "endowment:webassembly": {}
 },
 ```
+
+## Dynamic permissions
+
+Dynamic permissions are not requested in the manifest file. Instead, they are acquired during the lifecycle of the snap.
+
+## eth_accounts
```
Montoya, line 187: This should be 3 `#` not 2. So it indents correctly.

snaps/reference/permissions.md
Montoya: This is a technical reference that documents the permissions. A separate page explaining how to use `eth_requestAccounts` would provide a best practice.

Montoya (changes requested):
See requested change on line 187

hmalik88 (approved):

Montoya (approved):

alexandratran (approved):
Edited and made minor additions for clarity. LGTM.

FrederikBolding (approved):

PR #904, opened 2023-09-06
Update numerous outdated examples in Anatomy of a Snap

Montoya:


snaps/concepts/anatomy.md
```diff
@@ -157,14 +153,14 @@ details the requirements of both `snap.manifest.json` and its relationship to `p
...
-For example, if you change the location of the (optional) icon SVG file, you must update
+For example, if you change the location of the icon SVG file, you must update
 `source.location.npm.iconPath` to match.
 You can also use the [command line](../reference/cli/index.md) to update some fields for you.
 For example, `mm-snap build` or `mm-snap manifest --fix` updates `source.shasum`.
 
 ## Configuration file
 
-The snap configuration file, `snap.config.js`, should be placed in the project root directory.
+The snap configuration file, `snap.config.ts`, should be placed in the project root directory.
```
Mrtenz, line 163: Since you changed this, I would change the example below as well, to use ES6 import/export syntax.

We should also change this section to use Webpack, but out of scope for this PR.

Mrtenz (changes requested):

snaps/concepts/anatomy.md
Montoya: OK I have updated the example and removed the outdated browserify example below it to be rewritten later.

Mrtenz (approved):

PR #952, opened 2023-10-04
Remove communicationServerUrl option from React SDK example

alexandratran:
Remove unnecessary `communicationServerUrl` option from the React SDK example.


andreahaku (commented):
LGTM

PR #965, opened 2023-10-18
Add `snap_getLocale` documentation

ziad-saab:


ziad-saab:
@SocketSecurity ignore-all

snaps/index.mdx
```diff
@@ -78,6 +78,12 @@ Features include:
       href: "reference/permissions#endowmentnetwork-access",
       title: "Network access",
       description: <>Make API calls using <code>fetch()</code>.</>
+    },
+    {
+      href: "reference/rpc-api#snap_getlocale",
```
FrederikBolding, line 83: Needs an icon

cc @eriknson

package-lock.json
FrederikBolding, line 1: Is deleting this file intentional?

snaps/index.mdx
Montoya: He already identified the `home` icon from the DS to be used for the permission, so we just need to find that icon and use it here

Montoya: Just pushed an update with the correct icon!

src/theme/DocSidebarItem/index.module.css
```diff
@@ -2,15 +2,9 @@
...
   border-radius: 4px;
   padding: 2px 4px;
   transform: none;
-}
-
-html[data-theme='dark'] .flaskOnly > :global(a.menu__link::after),
-html[data-theme='dark'] .flaskOnly > :global(.menu__list-item-collapsible) > :global(a.menu__link::after) {
-  color: hsl(280, 45%, 85%);
-  background-color: hsl(280, 45%, 25%);
 }
```
Mrtenz, line 10:
```suggestion
}

```

Mrtenz (changes requested):

FrederikBolding (approved):

FrederikBolding (approved):

snaps/reference/rpc-api.md
```diff
@@ -562,6 +562,40 @@ console.log(entropy);
...
+### snap_getLocale
+
+:::flaskOnly
+:::
+
+Gets the user's locale setting. You can use this method to localize text in your snap.
+
+#### Returns
+
+The user's locale setting.
```
vandan, line 574: Should we provide detail on the specific locale format that's returned here? Is there a reference that can be included?

alexandratran: @ziad-saab Do we expect the return format to be an ISO two-letter language code? https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes

Montoya: Hadn't thought about this. What is the format that MetaMask uses?

Montoya: OK, regardless of format, here's the locales: https://github.com/MetaMask/metamask-extension/blob/develop/app/_locales/index.json

What's the best way you suggest surfacing this information in the doc? Would be happy with a link to this file.

alexandratran, line 574:
```suggestion
The user's locale setting as a [language code](https://github.com/MetaMask/metamask-extension/blob/develop/app/_locales/index.json).
```

Here's a suggestion @Montoya

alexandratran (approved):
LGTM

PR #991, opened 2023-11-13
Changed docusaurus-openrpc plugin to be latest version to get the index page feature

### LINKED ISSUE #912: Incorporate a base landing page into the JSON-RPC API Playground section
Guidance from @shanejonas :

> in docusaurus-openrpc repo render the `info.description` field as the index page

### LINKED ISSUE #990: API Reference Pages Not Indexed in Algolia Search
The API reference pages are not appearing in Algolia search results on the MetaMask documentation site, hindering the ability of developers to find necessary information quickly.

shanejonas:
this is to add the latest updates in docusaurus-openrpc that adds the index page support. 

from this PR: https://github.com/MetaMask/docusaurus-openrpc/pull/57

fixes #990
fixes #912


package.json
```diff
@@ -21,11 +21,11 @@
   "dependencies": {
     "@docusaurus/core": "2.4.3",
     "@docusaurus/plugin-content-docs": "2.4.3",
-    "@docusaurus/theme-common": "^2.4.3",
-    "@docusaurus/theme-mermaid": "^2.4.3",
+    "@docusaurus/theme-common": "2.4.3",
+    "@docusaurus/theme-mermaid": "2.4.3",
```
adonesky1, line 25: We don't minor version resolution dedupe bumps? I guess they probably wouldn't happen in this dep tree. What is the reason for this change just curious?

shanejonas: We need the docusaurus version to match the plugin versions so linking works properly with `docusaurus-openrpc`. and the docusaurus-openrpc has a dep on `@docusaurus/theme-common` as well.

adonesky1 (approved):

PR #999, opened 2023-11-16
Remove manual rpc reference

### LINKED ISSUE #816: port over docs from JSON-RPC API docs and remove the former.

alexandratran:
Remove manual RPC API reference, update links, add redirect. Fixes #816.


vandan:
With this change, I think we should call the section the _**JSON-RPC API Reference**_ and just note in other places that it's interactive.

vandan:
Not needed in this PR since it would go to the api-spec repo, but should we use this opportunity to incorporate the notes about JSON-RPC [API tags](https://docs.metamask.io/816-remove-manual-rpc-api/wallet/concepts/apis/#json-rpc-api) that are currently described under Concepts into the index page for the JSON-RPC API.

vandan (changes requested):
Looks like we still need to apply the change from "Playground" to "Reference" in the sidebar

vandan (approved):

PR #1000, opened 2023-11-16
update ios sdk docs to 0.3.0

alexandratran:
Update iOS SDK setup docs to version 0.3.0, using async/await instead of callbacks.

Preview: https://docs.metamask.io/ios-updates/wallet/how-to/connect/set-up-sdk/mobile/ios/


elefantel (approved):
Looks good

PR #1085, opened 2024-01-15
Fix docs for security guidelines

pkeshkamat:
Hey,
The code from the current guidelines:
```js
    const referrer = new URL(origin);

    if(referrer.protocol === "https:" && referrer.host.endsWith("metamask.io")) { 
      console.log("URL is valid"); 
    }
    else { 
      console.log("URL is NOT valid"); 
    }
```
will also end up matching the following URLs as valid URLs due to the functionality of .endsWith:
https://ametamask.io
https://abcdefgmetamask.io
The `.` is necessary before `metamask.io` to ensure the correct origin.

### DIFF
--- a/snaps/concepts/security-guidelines.md
+++ b/snaps/concepts/security-guidelines.md
@@ -75,7 +75,7 @@ The following are guidelines for user notifications and authorizations:
     ```JavaScript
     const referrer = new URL(origin);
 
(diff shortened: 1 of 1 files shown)


Montoya:
Excellent catch, thank you!

Montoya (approved):

GuiBibeau (approved):

NicolasMassart (approved):

PR #1111, opened 2024-01-27
Fixed: 1097 Clean up redirects

saakarx:
Fixed redirects from separate routes to single array of routes


alexandratran:
This looks great! However, the open PR #1114 required a lot of redirect updates, so to avoid nasty merge conflicts I have manually ported over these changes to #1114 and will close this in favor of that PR. Thanks for the contribution!

Issue #1133, opened 2024-02-13
All the carts should be of same height


subhamkumarr:
## Description:
The height of all the carts is not the same. All the carts should be the same.

## Screenshort:
![Screenshot (1098)](https://github.com/MetaMask/metamask-docs/assets/99977240/5c80545e-941b-4aad-9924-9a3ddebc4a3c)

subhamkumarr:
@alexandratran, may I work on this issue, ma'am?

alexandratran:
@subhamkumarr assigned to you. Are you still working on the other issue that you raised? (#1105). The discussion channel for MetaMask devs is in the [Consensys Discord](https://discord.gg/consensys).

subhamkumarr:
Yes, ma'am, I'm working on this issue https://github.com/MetaMask/metamask-docs/issues/1105. This week, I will raise a PR for both issues.

PR #1136, opened 2024-02-14
fix: #1133:  All the carts should be of same height

subhamkumarr:
## Description:
All the carts is having same height

#### Before:
Screenshot (1098)

#### After:
image

### DIFF
--- a/src/components/Card.tsx
+++ b/src/components/Card.tsx
@@ -12,7 +12,7 @@ export type CardItem = {
 export default function Card({ title, link, description }: CardItem) {
   return (
     <div className="col col--4 margin-top--sm margin-bottom--md">
-      <Link className={clsx(styles.root, "card padding--sm")} href={link}>
+      <Link className={clsx(styles.root, "card")} href={link}>
         <div className="card__header"><h3>{title}</h3></div>
         <div className="card__body">{description}</div>
       </Link>


subhamkumarr:
@alexandratran could you please review the PR

src/components/Card.tsx
```diff
@@ -12,7 +12,7 @@ export type CardItem = {
 export default function Card({ title, link, description }: CardItem) {
   return (
     <div className="col col--4 margin-top--sm margin-bottom--md">
-      <Link className={clsx(styles.root, "card padding--sm")} href={link}>
+      <Link className={clsx(styles.root, "card")} href={link}>
```
alexandratran, line 15: @subhamkumarr It looks like the homepage cards have small padding by default, so `padding--sm` is redundant and I can merge this change if you want. However, this doesn't quite fix the issue of cards being different heights – the card height still depends on the amount of text in the card + the width of your screen.

subhamkumarr: By fixing this even i observe that the card height depends on the amount of text in the card. I look into all other card this change does not harm other card. 

I want this to be merge. This change making home page look beautiful.

alexandratran (approved):

alexandratran:
@subhamkumarr Sorry, I do not help maintain the `metamask-extension` repo. Just give them a little more time or ping on the [Consensys Discord](https://discord.gg/consensys).

PR #1175, opened 2024-02-26
fix: snaps transaction-insights missing dependencies

marvinmarnold:
Missing @metamask/snaps-types and @metamask/snaps-ui which are referenced later in index.ts.

### DIFF
--- a/snaps/learn/tutorials/transaction-insights.md
+++ b/snaps/learn/tutorials/transaction-insights.md
@@ -49,7 +49,12 @@ Next, `cd` into the `transaction-insights-snap` project directory and run:
 yarn install
 ```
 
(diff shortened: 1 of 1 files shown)


joaniefromtheblock:
Thank you for this helpful PR @marvinmarnold and for catching this. We'll update the docs accordingly.

Montoya:
When using the latest template, the dependency is `@metamask/snaps-sdk`. The mentioned packages are deprecated in favor of this package. Please test the flow and see if it works with the latest template before making this change.

Montoya (changes requested):
See above comment

Montoya:
Closing in favor of #1206

PR #1204, opened 2024-03-10
Homepage and getLocale is no longer flask-only

Montoya:
This removes the flask-only admonition from the home page endowment and entry point.


alexandratran (commented):
On the Snaps introduction page, can you also move the "Home pages" card to the non-Flask card list? https://github.com/MetaMask/metamask-docs/blob/094b68c20a00dc5448b07bc4de612e8e09e8ba39/snaps/index.mdx#L120-L126

alexandratran (approved):

PR #1223, opened 2024-03-18
Add how-to on restricting rpc apis by URL

ziad-saab:
closes MetaMask/MetaMask-planning#1285.


Montoya:
I think the tutorial should use the URL library like so: https://codepen.io/m0nt0y4/pen/WNLxYGw

Relying on `Array.prototype.includes()` will miss if the origin is "https://metamask.io" and the array contains "https://metamask.io/" or if the origin is "https://metamask.io/dapp/" and the array contains "https://metamask.io". The point of using the URL library is the extract the most important parts of the origin (protocol + top level domain) and key based on that. Then you can just store the TLD in the array and compare against that, and require the request to always come from https. Or store both protocol + TLD in the array and compare against both.

Montoya (changes requested):
See above

snaps/how-to/restrict-rpc-api.md
```diff
@@ -0,0 +1,58 @@
...
+}) => {
+  // Check permissions
+  if (!isAllowed(request.method, origin)) {
+    throw new UnauthorizedError(`Method ${request.method} not authorized for origin ${origin}.`);
+  }
+
+  switch (request.method) {
+    case 'hello':
+      return 'world!';
+
```
Montoya, line 51: Is there supposed to be `case 'secureMethod':` here?

snaps/reference/permissions.md
```diff
@@ -216,6 +216,8 @@ Specify this permission in the manifest file as follows:
 }
 ```
 
+#### Allowed origins
+
 Alternatively, you can specify the caveat `allowedOrigins` to restrict requests to specific domains or Snap IDs.
```
Montoya, line 221: Maybe this should say "to restrict **all** requests" to differentiate from the guide we are adding

snaps/how-to/restrict-rpc-api.md
```diff
@@ -0,0 +1,58 @@
+---
+description: Restrict a Snap's RPC API methods
+sidebar_position: 7
+---
+
+# Restrict a Snap's RPC API methods
+
+## All Dapps are allowed by default
```
Montoya, line 8: Dapps should be lowercase like dapps

snaps/how-to/restrict-rpc-api.md
```diff
@@ -0,0 +1,58 @@
...
+
+By default, a Snap's RPC API is callable by any Dapp if the [`dapps` caveat to `endowment:rpc`](../reference/permissions.md#endowmentrpc) is set to `true` in the Snap's manifest.
```
Montoya, line 10: any **dapp**

Montoya (changes requested):

ziad-saab:
> Relying on `Array.prototype.includes()` will miss if the origin is "https://metamask.io" and the array contains "https://metamask.io/" or if the origin is "https://metamask.io/dapp/" and the array contains "https://metamask.io".

@Montoya it's my understanding and experience that the `origin` parameter is already just protocol + domain. it doesn't contain any path information. Do we still think it's necessary to include the URL manipulation in this case?

Montoya:
@ziad-saab Oh, maybe my understanding is wrong. If that is true then this approach should work fine.

alexandratran (approved):
Minor edits to content. LGTM.

Montoya (approved):

PR #1234, opened 2024-03-26
Update video formatting

### LINKED ISSUE #1072: Improve video formats
1. Videos being located in the `static` folder make them not show up in PR previews because the previews attach the PR branch name to the beginning of the URL. See if we can put videos in a different folder(s) instead so they work consistently in local and PR previews.

2. Also, some of the embedded YouTube videos don't fit on screen on mobile devices. For example, the Unity SDK demo:
  
    image
  
    Maybe use `YoutubeEmbed` that is used in the Snaps docs to embed these videos responsively.
    We should also enable users to play the videos full-screen without opening a new a tab.

3. Increase border radius for all videos and gifs, to match our image and code sample styles.

joaniefromtheblock:
Closes #1072


alexandratran (changes requested):
To close issue #1072, this PR still needs to address the first two points described in that issue:

1. Relocate videos in the `static` folder to fix the PR preview issue.
2. Fix the rendering of the Unity SDK video and make it consistent with the other YouTube embeds.

Please let me know if you have questions or blockers.

wallet/how-to/use-sdk/gaming/unity/index.md
```diff
@@ -20,7 +20,7 @@ The MetaMask Unity SDK supports macOS, Windows, Linux, iOS, Android, and WebGL.
 The following video demonstrates how to install and use the MetaMask Unity SDK.
 
 <p align="center">
-  <ReactPlayer url='https://www.youtube.com/embed/0D1cIH-PZtI' />
+  <YoutubeEmbed url='https://www.youtube.com/embed/0D1cIH-PZtI' />
```
alexandratran: This does not appear in the rendered preview and requires an import statement at the top of the page:

```
import YoutubeEmbed from '@site/src/components/YoutubeEmbed';
```

wallet/concepts/sdk/index.md
```diff
@@ -56,7 +56,7 @@ to MetaMask Mobile using a QR code.
 
 <p align="center">
   <video width="100%" controls>
-    <source src="/sdk-desktop.mp4" type="video/mp4" />
+    <source src="/public-videos/sdk-desktop.mp4" type="video/mp4" />
```
alexandratran, line 59: These links don't match the folder name you created `public_videos` and so don't render in the preview

alexandratran (changes requested):
Please always check previews locally and on the PR.

To match how we organize images, can we add a `wallet/assets/videos` folder and `snaps/assets/video` folder and add the corresponding videos in those folders? Then update the video src links and check that they render properly. I'm not sure if relative or absolute links are better to use in this case.

Montoya (approved):

PR #1238, opened 2024-03-29
Document using node.js modules in Snaps

### LINKED ISSUE #855: Document how to use native Node.js modules in snaps
The new Snaps CLI will show an error when the snap is trying to use native Node.js modules (`crypto`, `path`, etc.), or `Buffer`. We should add instructions on how to resolve these errors.

References:

- https://github.com/MetaMask/snaps/blob/678abd9a550140a47b38a38bce489e7375d8f92c/packages/snaps-cli/src/webpack/plugins.ts#L394-L406
- https://github.com/MetaMask/snaps/blob/678abd9a550140a47b38a38bce489e7375d8f92c/packages/snaps-cli/src/webpack/plugins.ts#L460-L469

alexandratran:
This PR slightly rearranges the "Snaps execution environment" concept page and adds a note on how to use the Node.js built-in modules, which are not supported by default. It also documents the missing `polyfills` config option. Fixes #855.

Preview: https://docs.metamask.io/855-nodejs-modules/snaps/learn/about-snaps/execution-environment/

### DIFF
--- a/snaps/learn/about-snaps/execution-environment.md
+++ b/snaps/learn/about-snaps/execution-environment.md
@@ -50,9 +50,8 @@ The following globals are also available:
 - `URL`
 
 :::info note
-To use the `Buffer` global, or a Node.js built-in module such as `crypto` or `path`, you
-must [provide a fallback](https://webpack.js.org/configuration/resolve/#resolvefallback) using the
-[`customizeWebpackConfig`](../../reference/cli/options.md#customizewebpackconfig) configuration option.
+To use Node.js built-in modules such as `crypto` and `path`, set the
+[`polyfills`](../../reference/cli/options.md#polyfills) option to `true`.
 :::
 
 ## Secure ECMAScript (SES)
--- a/snaps/learn/about-snaps/execution-environment.md
+++ b/snaps/learn/about-snaps/execution-environment.md
@@ -13,12 +13,23 @@ other parts of the application.
 This environment does not have a DOM, Node.js built-ins, or platform-specific APIs other than the
 default `snap` global and MetaMask's `ethereum` global.
 
-:::note
+The execution environment is designed to:
+
+- Prevent Snaps from influencing any other running code, including MetaMask itself.
+  That is, it prevents all Snaps from polluting the global environment and malicious Snaps from
+  stealing from users.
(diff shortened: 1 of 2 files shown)


Mrtenz (changes requested):
There is a `polyfills` option in the config which doesn't require changing the Webpack config. Looks like we forgot to add this to the CLI's README 😅

Mrtenz (approved):

snaps/reference/cli/options.md
```diff
@@ -390,6 +390,36 @@ output: {
...
+
+</TabItem>
+</Tabs>
+
+Enables or disables providing polyfills for Node.js built-in modules.
+If set to `true`, all available polyfills are provided.
+The default is `false`.
+
+You can also set this option to an object with specific polyfills set to `true`.
+See [the list of available polyfills](https://github.com/MetaMask/snaps/blob/51a1d04ea50c5c286262df1959ef0b1ced84b6e2/packages/snaps-cli/src/config.ts#L383-L416).
```
joaniefromtheblock, line 421: I think this is a fragment

```suggestion
For more information, see [the list of available polyfills](https://github.com/MetaMask/snaps/blob/51a1d04ea50c5c286262df1959ef0b1ced84b6e2/packages/snaps-cli/src/config.ts#L383-L416).
```

snaps/learn/about-snaps/execution-environment.md
```diff
@@ -13,12 +13,22 @@ other parts of the application.
...
 
-:::note
+The execution environment is designed to:
+
+- Prevent Snaps from polluting the global environment.
+- Prevent malicious Snaps from stealing from users.
+- Prevent Snaps from accessing sensitive JavaScript global APIs (such as `fetch`) without permission.
+- Be "fully virtualizable," or platform-independent.
+
+This allows you to safely execute Snaps anywhere, without the Snap needing to worry about where and
```
joaniefromtheblock, line 23: I think uses anthropomorphism (e.g. worry). Maybe the following instead?

```suggestion
This ensures that Snaps can be executed safely in any environment, without needing to consider their execution context or method.
```

joaniefromtheblock (approved):
LGTM

PR #1242, opened 2024-04-01
Clarify where to update Snap repository.url

### LINKED ISSUE #1168: Github Url in snap.manifest.json gets overwritten
When I change my Github Url in the `snap.manifest.json` file, and then attempt to build the Snap project, I notice that the URL gets overwritten with the Snap Template Github repo's Url.

On checking further, what's happening is that the Build script looks at the `package.json` file and pulls the Github URL from there and overwrites it in the `snap.manifest.json` file.
So the recommended approach would be to update the URL in the `package.json` file first, and then build the Snap project. Now, the build script copies over the URL from `package.json` and puts it in the `snap.manifest.json` file.

However, this causes some confusion for users who would not be aware of this "_undocumented_" feature and may assume that the snaps cli is broken.

This has to be remedied in a suitable way.
image

alexandratran:
This PR adds a recommendation to update `repository.url` and `version` in `package.json` instead of `snap.manifest.json`. Fixes #1168.

### DIFF
--- a/snaps/learn/about-snaps/files.md
+++ b/snaps/learn/about-snaps/files.md
@@ -75,23 +75,30 @@ The manifest tells MetaMask important information about your Snap, such as where
 reproduce the `source.shasum` value), and what
 [permissions the Snap requests](../../how-to/request-permissions.md) (using `initialPermissions`).
 
-:::note
-Currently, Snaps can only be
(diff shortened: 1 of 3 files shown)


mario-christopher (approved):

Montoya (approved):

bgravenorst (approved):
lgtm

PR #1287, opened 2024-04-23
Update design-guidelines.md

eriknson:
Minor snaps design guidelines update

### DIFF
--- a/snaps/learn/best-practices/design-guidelines.md
+++ b/snaps/learn/best-practices/design-guidelines.md
@@ -26,7 +26,7 @@ Ensure your Snap metadata follows these guidelines:
 - Keep your Snap's name to **21 characters or fewer** (including spaces).
 - Use a descriptive name&mdash;don't use "Snap" in the name.
 - Use short and simple copy.
-- Your avatar should fit in a **32px circular frame, SVG format**.
+- [Your Snap's avatar](https://docs.metamask.io/snaps/learn/tutorials/gas-estimation/#2-add-a-custom-icon) should fit in a **32px circular frame, SVG format**.
 
 **Before** asking for permission to install, provide users with clear and concise information about:
 


mario-christopher (approved):

FrederikBolding (approved):

alexandratran:
@eriknson Since we refer to it as an "icon" in the tutorial, which seems to be a more common term than "avatar," can we update the term to "icon" on this page? Could you also update the images `install.png` and `dialog.png` to use "icon" instead of "avatar"?

eriknson:
> @eriknson Since we refer to it as an "icon" in the tutorial, which seems to be a more common term than "avatar," can we update the term to "icon" on this page? Could you also update the images `install.png` and `dialog.png` to use "icon" instead of "avatar"?

No, the more common term for this context is avatar, so let's use that everywhere imo.

alexandratran:
@mario-christopher @eriknson Can we get agreement on avatar vs. icon? For me it seems like an [avatar](https://en.wikipedia.org/wiki/Avatar_(computing)) is a type of icon typically representing a user, not a brand or application. But I don't have a strong opinion on this.

And if we change this in the Snaps docs, should we also change the [Wallet docs](https://docs.metamask.io/wallet/how-to/display/icon/) to "display a dapp avatar"?

eriknson:
@alexandratran 

I just ran this by @Montoya and we agreed to use `Snap icon` or just `icon` (if it's already been introduced on the same page) so nvm above re: Avatar

alexandratran:
@eriknson Sounds good. Can you update the `install.png` and `dialog.png` images in [this section of the design guidelines](https://docs.metamask.io/snaps/learn/best-practices/design-guidelines/#optimize-your-metadata) to use "icon" instead of "avatar," or provide access to the raw files so I can update them? I'll go ahead and update the rest of the page.

alexandratran (approved):

PR #1314, opened 2024-05-20
chore: reword

jxom:
Rewording to more accurately phrase limitations of relying on a continuous connection from a mobile browser to a mobile app, rather than it being an issue of the upstream provider.

### DIFF
--- a/wallet/how-to/use-sdk/3rd-party-libraries/wagmi.md
+++ b/wallet/how-to/use-sdk/3rd-party-libraries/wagmi.md
@@ -110,7 +110,8 @@ write();
 
 ## Benefits of using the Infura API with Wagmi
 
-Mobile dapps might lose their continuous connection with MetaMask, causing read-only requests to fail.
+Read-only requests are blockchain requests that do not require user wallet interaction.
+Mobile dapps can lose their continuous connection with MetaMask, causing read-only requests to fail.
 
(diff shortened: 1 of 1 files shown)


jxom:
Why “when using Wagmi”? This puts a negative connotation onto the library when it’s not even our responsibility. 😅 It is also evident with other libraries such as Ethers & Web3. The MetaMask SDK should maintain the persistent connection even when the app is in the background.

wallet/how-to/use-sdk/3rd-party-libraries/wagmi.md
```diff
@@ -110,18 +110,14 @@ write();
 
 ## Benefits of using the Infura API with Wagmi
 
-Wagmi is not optimized for mobile environments.
-This limitation becomes evident when making read-only requests, which are queries that fetch data
-from the blockchain without making a transaction.
-Since mobile dapps might not maintain a continuous connection to MetaMask, these read-only requests
-can fail, leading to a suboptimal user experience.
+Mobile dapps might lose their continuous connection with MetaMask, causing read-only requests to fail.
```
alexandratran, line 113:
```suggestion
Read-only requests are blockchain requests that do not require user wallet interaction.
Mobile dapps can lose their continuous connection with MetaMask, causing read-only requests to fail.
```

alexandratran (approved):

PR #1325, opened 2024-05-31
Test adding Infura to the MetaMask docs

bgravenorst:
Test adding Infura to the MetaMask docs


services/concepts/failover-protection.md
```diff
@@ -0,0 +1,43 @@
...
+
+Failover protection is available on the Mainnet network for customers on the Growth or Custom plans.
+This feature ensures API requests are fulfilled even if Infura experiences a service issue or
+outage. This is achieved by routing requests to a select partner.
+
+Failover protection is enabled on a per-request basis by adding the failover header to your API
+request (see [Enable API request forwarding](../how-to/enable-api-forwarding.md)). This gives
+you control over which requests you'd like fulfilled by Infura's partner if failover protection should be needed.
+
+## Partners and privacy policies
```
bgravenorst, line 22: As a general topic this needs updating. For example, we'll need to list which networks supports Failover.
Also, different partners support different failover networks.

joaniefromtheblock: Thank you @bgravenorst. There are several sections where the direct transfer of information from the Infura docs to MetaMask requires some refinement (or simply doesn't fit well). I started a Google doc when I was doing the restructure, to identify content that may need some "MetaMasking". Maybe we can review it as a team and potentially create some tickets to tackle issues in the future.  For instance, somethings I mention are related to: 

- How we present Infura within MetaMask (RPC nodes for MM)
- Which tutorials are relevant to a user (I'm not sure, but some of the Infura tutorials look like they don't have much to do with MM)
- Where and when to add  testnet documentation (it's presently omitted in MM etc. 

I think we will just iteratively work on it, once the initial doc set is merged.

services/concepts/failover-protection.md
```diff
@@ -0,0 +1,43 @@
+---
+description: Learn about failover protection
+sidebar_position: 1
+---
+
+# Failover protection
+
+:::info
+
+Failover support is available on Mainnet only.
```
bgravenorst, line 10: This is currently incorrect and misleading so should be updated before merging. We should list the correct networks that support failover. Additionally, perhaps move the "Partners and privacy policies" content to the "services/reference/[NETWORK]/index.md file for the appropriate networks.

Note: Networks that support failover have different partners so it cannot be generic.

services/reference/ipfs/index.md
```diff
@@ -0,0 +1,27 @@
+---
+description: Gas network documentation
```
bgravenorst, line 2: Missing the network description

bgravenorst: I meant to say `description` is incorrect, and the topic requires a description for what IPFS is/does etc.

m4sterbunny (dismissed):
as per @bgravenorst approval in Slack

m4sterbunny (approved):
all tests pass

package.json
```diff
@@ -34,7 +34,8 @@
     "react-dom": "^18.0.0",
     "react-player": "^2.13.0",
     "remark-codesandbox": "^0.10.1",
-    "remark-docusaurus-tabs": "^0.2.0"
```
Rahul-cyber26: #1877

PR #1360, opened 2024-06-21
feat: ACT-1359 - Add Interactive Params Table

aednikanov:
# Description

Add an interactive params table and handle simple types.

## Screenshot

Screenshot 2024-06-21 at 11 03 16


TrofimovAnton85 (dismissed):

alexandratran (commented):
1. Would it make sense to make the value boxes a slightly different background color to make it clear that they are the only clickable/editable boxes?

    Screenshot 2024-06-21 at 9 54 33 AM

2. IDK if this was added in this PR, but the error table headers are the wrong font color in light mode:

    Screenshot 2024-06-21 at 9 49 36 AM

aednikanov:
@alexandratran it's implementation just like in the design: https://www.figma.com/design/gBbfVUyOdjS5OYZcg9mbUX/Sandbox-%2F-Interactive-API?node-id=492-21938&t=T8XsKFHK3b2e75ld-0

1. It's the approved design. You could tag Katie and leave your proposal if you want. It's nice to collect feedback from different people with their views and suggestions. And then we will have a better version!
But in this case, I think we should go forward, and don't stop on small things. It could be fixed quickly then if needed. Thanks.

2. Good catch! It was working on localhost but for some reason, the priorities of styles for the error table header were not the same after deployment. I reorganized little few styles, and now it works in both cases.

Also, could you, please, take a look at these alerts?
What could we do to pass or ignore them? I need `[json-schema-ref-parser@11.6.4](https://socket.dev/npm/package/@apidevtools/json-schema-ref-parser/overview/11.6.4)` for sure, it's the main tool for working with the JSON schema.

https://github.com/MetaMask/metamask-docs/pull/1360/checks?check_run_id=26636839795

alexandratran (approved):
LGTM! You can merge the PR to ignore the Socket Security check.

aednikanov:
@alexandratran thanks!

PR #1389, opened 2024-07-05
feat(reference): act-1327 - added support eip-6963

TrofimovAnton85:
## Description:

- Added support **EIP-6963** ( detect if the user has both `MetaMask` and `MetaMask Flask` installed on their browser and if this is the case propose them to pick the wallet to choose to interact with the MetaMask Docs )

##  How to test:

- Go to the test page - **/wallet/reference/new-reference**
- Test what you can choose from two providers (install them first)
Screenshot 2024-07-05 at 17 28 10

## Links:

[JIRA Issue](https://infura.atlassian.net/browse/ACT-1327)


alexandratran (approved):
LGTM

TrofimovAnton85:
> LGTM

Hey @alexandratran ! Could you please re-approve
I committed new changes (it turned out that metamask and metamask flask are essentially the same provider and they open at login at the same time, so we removed the choice between them).

alexandratran (approved):

PR #1394, opened 2024-07-08
feat(reference): act-1447 - added capture events to reference pages

TrofimovAnton85:
## Description

- added capture events ( "`Page Viewed`", "`Request Sent`", "`Request Configuration Started`",  "`Customize Request`" ) to new reference pages

Screenshot 2024-07-08 at 15 00 34
Screenshot 2024-07-08 at 15 01 56
Screenshot 2024-07-08 at 15 03 29

## Links:

[JIRA Issue](https://infura.atlassian.net/browse/ACT-1447)

### DIFF
--- a/src/lib/segmentAnalytics.js
+++ b/src/lib/segmentAnalytics.js
@@ -1,25 +1,45 @@
-export const trackClickForSegmentAnalytics = ({
-  type,
-  clickUrl,
-  clickText,
-  location,
-  userId,
+export const trackPageViewForSegment = ({
+  name,
+  path,
+  userExperience
 }) => {
   if (window.analytics) {
-    window.analytics.track(`CTA ${type} Clicked`, {
-      ...(clickUrl && { click_url: clickUrl }),
-      ...(clickText && { click_text: clickText }),
-      ...(location && { location }),
-      ...(userId && { user_id: userId }),
+    window.analytics.page("Page viewed", name, {
+      ...(path && { path: path }),
+      ...(userExperience && { user_experience: userExperience })
     });
   }
 };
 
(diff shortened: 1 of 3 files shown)


src/lib/segmentAnalytics.js
```diff
@@ -1,25 +1,45 @@
...
+      ...(userExperience && { user_experience: userExperience })
     });
   }
 };
 
-export const trackPageForSegmentAnalytics = ({ name, path, userId }) => {
+export const trackClickForSegment = ({
+  eventName,
+  clickType,
+  userExperience,
```
joaniefromtheblock, line 17: @TrofimovAnton85 Just out of curiousity what info does this provide in the analytics. Does it give us insight into loading metrics, or customer satisfaction, etc...?

TrofimovAnton85: @joaniefromtheblock  Hey! 
We have several types of events/metrics (they are described in the ticket) that we would like to collect for analytics. Based on them, we can test: what percentage of users use our updated reference pages and how convenient they are for them (do they use the parameter configurator, sending requests).

joaniefromtheblock (approved):
LGTM

src/lib/segmentAnalytics.js
joaniefromtheblock: sorry I meant specifically the `userExperience`?

TrofimovAnton85: First, we plan to release updated pages for 10% of users (based on the feature flag condition). For them, this `userExperience` metric will take the value “new”, and the remaining 90% users => old pages will be displayed.

aednikanov (approved):

PR #1395, opened 2024-07-08
feat(reference): ACT-1471 - Add UI Enhancements

aednikanov:
# Description

Add UI Enhancements according to the ticket (description only, from the comments we will create a new ticket): https://infura.atlassian.net/browse/ACT-1471


alexandratran:
@aednikanov @TrofimovAnton85 ^ It looks like this issue exists prior to this PR. It can be addressed separately.

alexandratran (approved):

aednikanov:
@alexandratran thanks. Yeah, we will have a ticket for managing form with keyboard controls. Because the current state regarding this is not ideal, as you see.

TrofimovAnton85 (approved):

PR #1424, opened 2024-07-18
feat(docs): added feature preview for ld

TrofimovAnton85:
## Description

-  added feature preview for ld

## Test:

- go to page **/wallet/reference/web3_clientversion**
- open WEB inspector in your browser and the **application** tab
- manually set the `feature_preview` cookie with the value `mmref`
- refresh the page and check that the new layout has loaded

Screenshot 2024-07-18 at 14 38 46

### DIFF
--- a/src/plugins/launchdarkly/ldClient.ts
+++ b/src/plugins/launchdarkly/ldClient.ts
@@ -9,10 +9,18 @@ export default (function () {
 
   const { LD_CLIENT_ID } = siteConfig.customFields;
 
+  const getCookie = (name: string) => {
+    let matches = document.cookie.match(new RegExp(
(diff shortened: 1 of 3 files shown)


src/plugins/launchdarkly/ldClient.ts
```diff
@@ -9,10 +9,18 @@ export default (function () {
 
   const { LD_CLIENT_ID } = siteConfig.customFields;
 
+  const getCookie = (name: string) => {
```
cedricmagne, line 12: To read the cookie use `js-cookie` instead of the regex

cedricmagne (changes requested):

src/plugins/launchdarkly/ldClient.ts
TrofimovAnton85: Good catch! Fixed & commited!

joaniefromtheblock (approved):
LGTM

PR #1448, opened 2024-07-24
ACT-1473 Assemble faucet page

Voltmod:
ACT-1473 Assemble faucet page


src/components/Input/input.module.scss
```diff
@@ -0,0 +1,72 @@
...
+    line-height: 22px;
+    outline: none;
+    width: 100%;
+    margin-bottom: 4px;
+
+    &:hover {
+    }
+
+    &:active {
+    }
```
aednikanov, line 51: Not needed code

PR #1450, opened 2024-07-25
fix(docs): removed init useeffect hook

TrofimovAnton85:
# Description

- Removed init `useEffect` hook to fix init detecting metamask extension.


Montoya (approved):
Checked the preview, works as expected!

joaniefromtheblock (approved):
LGTM. 

Previewed and I'm prompted to connect when I hit send request. I think this is the intended behavior.

PR #1579, opened 2024-09-26
feat(ref): act-1538 - added response example to ref pages

TrofimovAnton85:
## Description

- Added Response Example to the Reference Pages

## Test

- Go to the wallet ref page `/wallet/reference/eth_chainid/` and check that the right column displays the default response example 
- Click "Run request" button and check that the response has changed 
- Click "Response reset" (reset icon) and make sure that the default example has returned

Screenshot 2024-09-26 at 15 35 44

https://infura.atlassian.net/browse/ACT-1538


cedricmagne (approved):
Yeah it works.
Screenshot 2024-09-26 at 15 44 05
Screenshot 2024-09-26 at 15 44 25
Just for my understanding why the Connect wallet button is not hidden because when running the request we are using the chainId connected with Metamask

PR #1632, opened 2024-10-14
feat(ref): added linea reference feehistory page

TrofimovAnton85:
## Description

- Updated `eth_feehistory` reference page with dynamic component `<ParserOpenRPC />`
- Added fixes for init values ​​for array in the parameters table
- Refactored enum parameter description
- Set prod value for the endpoint of linea-mainnet network

## Test

- Go to page **services/reference/linea/json-rpc-methods/eth_feehistory/**
- Check default values in the params table (click "customize request" btn)
- Copy the request (in the section example request) and paste it into the terminal (**don't forget to substitute your real prod API key**) and execute the request

Screenshot 2024-10-14 at 13 32 00

### DIFF
--- a/services/reference/linea/json-rpc-methods/eth_feehistory.mdx
+++ b/services/reference/linea/json-rpc-methods/eth_feehistory.mdx
@@ -1,40 +1,13 @@
 ---
 title: "eth_feeHistory"
+hide_title: true
+hide_table_of_contents: true
 ---
 
-import Tabs from "@theme/Tabs"
-import TabItem from "@theme/TabItem"
+import ParserOpenRPC from "@site/src/components/ParserOpenRPC"
+import { NETWORK_NAMES } from "@site/src/plugins/plugin-json-rpc"
 
-import Description from "/services/reference/_partials/_eth_feehistory-description.mdx"
-
-<Description />
-
-## Parameters
-
-import Params from "/services/reference/_partials/_eth_feehistory-parameters.mdx"
-
-<Params />
-
-## Returns
-
-import Returns from "/services/reference/_partials/_eth_feehistory-returns.mdx"
-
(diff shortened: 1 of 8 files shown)


m4sterbunny (changes requested):
Looks great. Could we get the following fixes in with this PR?
- *Summary* (placeholder value of "Summary" in bold to anchor the summary data -- until design give alt guidance)
image
- Markdown links, AFAIK, the normal behavior is that external links open in new tab, while local links open in current (I am happy to be corrected that this isn't so)
image

TrofimovAnton85:
> Looks great. Could we get the following fixes in with this PR?
> 
> * _Summary_ (placeholder value of "Summary" in bold to anchor the summary data -- until design give alt guidance)
> 
> image
> - Markdown links, AFAIK, the normal behavior is that external links open in new tab, while local links open in current (I am happy to be corrected that this isn't so) image
> Probably worth brining the copy/reset into the top RHS of the box to allow the icons to be clearer/not obscured by the text below? image

---
Thanks!
1) Label "summary" => done!
2) Check for external links and open in a new tab => done!
3) About "copy button" - this is part of the internal docusaurus component `CodeBlock` and at the moment there is no way to change the behavior/styles

cedricmagne (approved):
Looks good.
I just wanted to notice that, I received `invalid project id` on all project API Keys after a curl in my terminal.

Also, agree with moving copy/past button to the header card instead, if it possible
Screenshot 2024-10-14 at 15 37 37

aednikanov (dismissed):

m4sterbunny (approved):
Still LGTM

cedricmagne (approved):

aednikanov (approved):

PR #1639, opened 2024-10-16
Recommend 3rd-party libraries

alexandratran:
# Description

Update wording on intro/connect pages to recommend third-party libraries for multi-wallet support.

## Preview

https://metamask-docs-6506mlvs9-metamask-web.vercel.app/wallet/

## Checklist


joaniefromtheblock (approved):
LGTM. Apologies. I am not sure why the Trivy/Run check is failing

httpJunkie (approved):
I fell that the updates are fine and do clarify our current suggestions for connecting to multiple wallets. I can see this page changing in the future if we have our own multi-wallet connector. But it looks good text-wise. I worry that having similar information about third part libraries on both pages is actually ok in it's current form as we briefly mention it on the introduction and cover it a bit more on the connection pages.

I like how we have the note about SDK only detecting MetaMask and then we go directly into the Third Party libs page.

PR #1719, opened 2024-11-07
Remove POW methods.

bgravenorst:
# Description

Removed the POW-related Linea methods. In some cases like `eth_getWork`, `eth_submitWork`, or `eth_protocolVersion` the methods display a `the method [methodName] does not exist/is not available` error

However some like the `eth_getUncle*` methods or `eth_hashRate` method will run but not return any logical result.

See: https://github.com/Consensys/butterfly/pull/17/files

## Checklist

- [x] If this PR contains a major change to the documentation content, I have added an entry to the top of the ["What's new?"](https://github.com/MetaMask/metamask-docs/blob/main/docs/whats-new.md) page.
- [x] The proposed changes have been reviewed and approved by a member of the documentation team.


alexandratran (commented):
Should we also remove `eth_accounts`?

bgravenorst:
> Should we also remove `eth_accounts`?

I think we need some clarification on some of these methods (perhaps as part of a future PR). `eth_accounts` is not a PoW-only function so we need to figure out how we document this in the future.

alexandratran (approved):

PR #1756, opened 2024-11-29
fix(faucet): hide login flow for mobile devices

TrofimovAnton85:
## Description

- hidden login flow for mobile devices and added the corresponding message

## Test

- on the desktop, go to the **/developer-tools/faucet/** and check the login button is available in the navigation bar and on the page
- check this page on a mobile device (or emulation mode in the web inspector - in this case, reload the page) and check that the button is hidden and there is a message

Screenshot 2024-11-29 at 15 19 22

Screenshot 2024-11-29 at 15 17 38

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,6 +42,7 @@
         "prism-react-renderer": "^2.1.0",
         "react": "^18.0.0",
         "react-alert": "^7.0.3",
+        "react-device-detect": "^2.2.3",
         "react-dom": "^18.0.0",
         "react-dropdown-select": "^4.11.2",
         "react-modal": "^3.16.1",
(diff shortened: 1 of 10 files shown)


AndyMBridges:
Hey @TrofimovAnton85 - Instead of using `react-device-detect` can we just hook into the `isExtensionActive` - if false, then display the message?

AndyMBridges (changes requested):
Hey @TrofimovAnton85 - Instead of using react-device-detect can we just hook into the isExtensionActive - if false, then display the message?

TrofimovAnton85:
> Hey @TrofimovAnton85 - Instead of importing `react-device-detect`, can we use any of our existing logic?

Hey! Thanks!
Fixed & committed!

PR #1818, opened 2025-01-18
Update link format 

Vikt0rPavlik:
Update link format in run-devnet.md

Changed in wallet/how-to/run-devnet.md:
Old: [nonce] (send-transactions.md#nonce)
New: [nonce] (/send-transactions.md#nonce)

Added leading forward slash to fix internal documentation link path and maintain consistency with MetaMask's documentation system.


alexandratran:
Relative links don't need a leading slash when you are linking to a file in the same directory. This link is working correctly already.

PR #1825, opened 2025-01-20
Fix testnet name: 'Polygon Amoy' → 'Polygon Mumbai'

Guayaba221:
Because the official Polygon testnet is called “Mumbai,” not “Amoy.”


bgravenorst:
Thanks for your contribution, however Polygon Mumbai was deprecated last year, see https://polygon.technology/blog/introducing-the-amoy-testnet-for-polygon-pos.

PR #2096, opened 2025-06-18
Ensure descriptions appear correctly.

### LINKED ISSUE #2089: Fix Description handling in OpenRPC component
Description handling needs to be updated to handle multiple levels in the spec.
For example, descriptions exist for [`eth_feeHistory`](https://docs.metamask.io/wallet/reference/json-rpc-methods/eth_feehistory/) but are not being displayed.

bgravenorst:
# Description
Preserve parameter descriptions when resolving schema references

## Problem

Parameter descriptions were not displaying correctly when parameters used schema references (`$ref`). Instead of showing the parameter-specific descriptions, the parser was displaying the generic schema titles from the referenced components.

## Root Cause

The `resolveRef` function in `RenderParams.tsx` was losing the original parameter's description when resolving schema references. When a parameter had a `$ref` schema, the function would:

1. Resolve the reference to get the target schema
2. Call `renderSchema` with only the resolved schema
3. **Lose the original parameter's description** in the process

## Solution

Modified the `resolveRef` function to preserve the original parameter's description when resolving references:

## Testing
Ensure descriptions are now displaying for methods like `eth_decrypt`, `eth_feeHistory` `eth_getEncryptionPublicKey`.
Ensure we still fallback to the types `title` when descriptions are absent.

## Fixes
Fixes #2089 

## Checklist

- [x] If this PR contains a major change to the documentation content, I have added an entry to the top of the ["What's new?"](https://github.com/MetaMask/metamask-docs/blob/main/docs/whats-new.md) page.
- [x] The proposed changes have been reviewed and approved by a member of the documentation team.


alexandratran (approved):
LGTM

PR #2103, opened 2025-06-18
Upgrade Docusaurus and minor fixes

alexandratran:
# Description

- Upgrade Docusaurus
- Ignore recurring lint code base error
- Update DTK tagline
- Add scroll padding so sticky product banner doesn't obscure headings
- Fix scroll issue for anchor links
- Fix CSS selector for announcement bar

## Fixes
#2092 

## Preview

https://metamask-docs-3tssqpuip-consensys-ddffed67.vercel.app/

## Checklist

Complete this checklist before merging your PR:

- [ ] If this PR contains a major change to the documentation content, I have added an entry to the top of the ["What's new?"](https://github.com/MetaMask/metamask-docs/blob/main/docs/whats-new.md) page.
- [ ] The proposed changes have been reviewed and approved by a member of the documentation team.

PR #2104, opened 2025-06-18
Fix scrolling issue breaking navigation on certain pages.

bgravenorst:
# Description

The right sidebar Table of Contents (TOC) links were not working properly on some pages. When users clicked on TOC links, the page would scroll to the correct section but then immediately jump back to the top, making navigation unusable.

## Root Cause
The issue was in `src/client/scroll-fix.js` where an overly aggressive `window.scrollTo(0, 0)` call was resetting the page scroll position after every route change. This interfered with:

- TOC anchor link navigation
- Direct anchor link navigation  
- User's intended scroll position

## Solution
Removed the `window.scrollTo(0, 0)` call from the scroll-fix logic while preserving the intended sidebar visibility functionality.

## Checklist


bgravenorst:
Fixed in #2103

PR #2158, opened 2025-07-18
fix `to` field in redeem delegations guide

### LINKED ISSUE #2082: to field in sendUserOperation() is incorrect in Quick Start example
## Description:

In the **Quick Start** section of the MetaMask Delegation Toolkit docs, the `sendUserOperation()` call sets the `to` field to the delegate's smart account address:

```js
  to: delegateSmartAccount.address
```

However, this is **incorrect** for a `redeemDelegation()` call. The correct `to` value should be the **delegator's smart account address** — the account that originally signed the delegation and will validate the `UserOperation` via `validateUserOp`.

This is properly shown later in the documentation under the **Redeem a delegation** section:

```js
  to: "<DELEGATOR-SMART-ACCOUNT-ADDRESS>"
```

## Issue:

Using the delegate's address causes the `UserOperation` to fail, often with no clear error unless debugged deeply. The delegation signature isn't validated on the delegate’s account, which leads to invalid operations.

## Fix:

Update the **Quick Start** example to use the **delegator's** address:

```js
  to: delegatorSmartAccount.address
```

## Impact:

Anyone following the **Quick Start** example verbatim will encounter failures during broadcast, especially when using `redeemDelegations()` logic. This discrepancy adds friction for developers and could be avoided with a small correction.

alexandratran:
# Description

Fix `to` field in the Redeem delegations guide to make it consistent with the quickstart.

## Issue(s) fixed

Fixes #2082 

## Checklist


delegation-toolkit/get-started/delegation-quickstart.md
```diff
@@ -146,7 +146,7 @@ const userOperationHash = await bundlerClient.sendUserOperation({
   account: delegateSmartAccount,
   calls: [
     {
-      to: delegateSmartAccount.address,
+      to: delegatorSmartAccount.address,
```
AyushBherwani1998: I'm not actually sure if we should document delegatorSmartAccount. It assumes the delegatorSmartAccount will be present always which is not the case, if the delegations are stored, or if delegations are used in different environment to redeem like server, it won't stand true. Hence it was using delegateSmartAccount.address. Maybe we should change the documentation for redeemDelegation rather than quickstart. WDYT?

alexandratran: If we keep `delegateSmartAccount` here, what about the issue stated in #2082?
> Using the delegate's address causes the `UserOperation` to fail, often with no clear error unless debugged deeply. The delegation signature isn't validated on the delegate’s account, which leads to invalid operations.

AyushBherwani1998: It should work fine, when `redeemDelegations` is called on Smart Account, it just works as a proxy for delegation manager. So all the calls are passed and processed by Delegation manager.

alexandratran: Ok, I've switched to updating the Redeem delegation guide instead, lmk if this is correct.

AyushBherwani1998 (approved):
LGTM

PR #3056, opened 2026-09-14
Give each chain its own social banner

yashovardhan:
## Summary

Chain pages under `connect-blockchain/` set an `image:` frontmatter that Docusaurus emits as `og:image`, so a wrong or missing value is exactly what people see when they share a link. An audit of every page in that tree turned up four distinct problems:

- **Wrong chain.** Arc and Monad were each copied from the Ethereum page and kept Ethereum's banner, across all of their pages. The live Arc page still serves `banners/ethereum.png`.
- **Missing file.** Seven chains referenced a banner that was never added, so the preview 404s: `base`, `bnb`, `astar-zkevm`, `astar-zkyoto`, `neon`, `nibiru`, `opbnb`.
- **Stale path.** `other/ton.mdx` pointed at a `guides/` prefix that doesn't resolve.
- **Inconsistent.** The Unity pages for Ethereum and Solana used the generic Unity banner, while all 44 other chains use their own chain banner there.

Every page under `connect-blockchain/` now resolves to a banner that exists and belongs to that chain.

## How the new banners were made

The eight new PNGs match the existing 756x350 template rather than approximating it. The template was recovered from the 76 banners already in the repo: a horizontal sRGB gradient from `#68A2FF` to a chain accent, the Web3Auth mark in a 144px disc at (247, 175), the MetaMask wordmark at x 337-418, and the chain logo in a 144px disc at (509, 175).

Regressing each pixel against its background gradient across all existing banners separates the fixed chrome's coverage from its colour exactly. The accent rule fell out of the same data: keep the logo's hue and saturation, lift lightness to about 0.70. Regenerating already-shipped banners from scratch reproduces them to within roughly 3/255 mean per-pixel error, which is the evidence that the reconstruction is faithful rather than merely close.

Chain logos come from `static/logos/`. Monad's only exists as SVG, so its path was flattened to a bitmap.

## Two judgement calls worth reviewing

- **opBNB** has no mark of its own anywhere in the repo. Its pages share the BNB banner. That is BNB Chain's mark rather than opBNB's, so it is accurate as to ecosystem but not as to the specific network. It replaces a state where the link had no chain branding at all.
- **astar-zkyoto** reuses the Astar zkEVM mark, since zKyoto is that network's testnet.

Both should be swapped out if design supplies dedicated artwork.

## Test plan

- [ ] Preview locally with `npm start` and confirm the new banners render at the expected size and contrast
- [ ] Spot-check `og:image` in the built HTML for `evm/arc/`, `evm/monad/`, `evm/base/`, `evm/opbnb/`, and `other/ton/`
- [ ] Confirm `npm run format:check` passes (verified locally on all 32 changed Markdown files)
- [ ] Design review of the eight generated banners, in particular opBNB and astar-zkyoto

Made with [Cursor](https://cursor.com)


shahbaz17 (approved):
lgtm

